PR #205, opened 2018-01-26
Hey Cameron. Again, not a JS expert, so if there are big-ticket issue…

partnerinflight:
…s here of course lmk and I'll work to fix it.

I thought I'd separate out the recent file handling code into a separate class -- S in SOLID, as much as possible. ;)

Lmk what you think.


bin/recentfilesmanager.js
```diff
@@ -0,0 +1,122 @@
+var Rollbar = require('rollbar')
+var storage = require('electron-json-storage')
+var log = require('electron-log')
+var TRIALMODE = process.env.TRIALMODE === 'true'
+
+var recentFiles = null
+
+const recentKey = process.env.NODE_ENV === 'dev' ? 'recentFilesDev' : 'recentFiles'
+
+module.exports = class RecentFilesManager {
```
cameronsutter: I don't think `class` will work here (if it does, then no problem) because i don't think this file is being compiled by babel (that's how i've set this project up: the renderer process is compiled, but the main process is not … i know it's not the best)

bin/recentfilesmanager.js
```diff
@@ -0,0 +1,122 @@
...
+
+  constructor() {
```
cameronsutter: no need to leave this in

bin/recentfilesmanager.js
```diff
@@ -0,0 +1,122 @@
...
+      return new Promise(function(resolve, reject) {
+        storage.has(recentKey, function(err, hasKey) {
+           if (err) { 
+             log.warn(err)
+             self.rollbar.warn(err)
+             resolve([])
+             return
+            }
+            if (hasKey) {
+            storage.get(recentKey, function (err, contents) {
```
cameronsutter: it looks like the indenting got a little off here

bin/recentfilesmanager.js
```diff
@@ -0,0 +1,122 @@
...
+const recentKey = process.env.NODE_ENV === 'dev' ? 'recentFilesDev' : 'recentFiles'
+
+module.exports = class RecentFilesManager {
+
+  constructor() {
+  }
+  
+  // returns the recents list. Note we never want to error out from here -- if there's a problem with the recents list, 
+  // just return an empty array
+  get recents() {
```
cameronsutter: I've never seen this `get` in js. Is that a thing?

bin/recentfilesmanager.js
```diff
@@ -0,0 +1,122 @@
...
+  
+        let result = [fileName];
+        await self.saveRecentsList(result);
+      })
+    });
+  }
+  
+async removeRecentFile (fileNameToRemove) {
+  log.info("Removing " + fileNameToRemove + " from recent files list")
+  let recentsList = await this.recents
```
cameronsutter: if my comment about `get recents()` being a thing above is wrong, then this should be fine. Otherwise it would have to be a `this.recents()`

cameronsutter:
going to close this because mac and windows does this for us now … sorry it's taken so long

PR #431, opened 2020-04-29
Optimize timeline

cameronsutter:


cameronsutter:
merged into master manually

PR #449, opened 2020-11-03
Convert unicode characters for rendering in RTF

shichongrui:
refs: BUG-60, BUG-58

It would be good to test this in Scrivener instead of just
a document viewer like Word

### DIFF
--- a/src/common/exporter/scrivener/v2/exporter.js
+++ b/src/common/exporter/scrivener/v2/exporter.js
@@ -110,6 +110,20 @@ function createScrivx (state, basePath) {
   return sceneCardsByDocID
 }
 
+function convertUnicode(str) {
+  let converted = '';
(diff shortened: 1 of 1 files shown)


src/common/exporter/scrivener/v2/exporter.js
```diff
@@ -130,8 +144,8 @@ function createDocuments (sceneCardsByDocID, basePath) {
     try {
       serialize(card.description, doc)
       data = doc.createDocument()
-      // may have to do this … but it doesn't look like it
-      // const buffer = new Buffer(data, 'binary')
+      data = convertUnicode(data)
+      data = Buffer.from(convertUnicode(data), 'utf8')
```
cameronsutter, line 148: You have to convertUnicode twice here? Once on each line?

cameronsutter (approved):

PR #456, opened 2020-11-12
Upgrade electron

shichongrui:
refs: BAC-37

Tested out several parts of the app such as opening files and saving. Opening other windows etc... One weird thing I noticed was that when I did File -> Open the first time, it focused the file explorer and let me choose. But if I did it a second time it didn't focus the file explorer and just showed me an OS level notification that Plottr was ready. And then I saw that there was a file explorer window open that I could use to open a new Plottr file. I'm not sure if this is an Ubuntu thing though so I figured it would be good to test on Mac/Windows.


cameronsutter:
I'm waiting to merge this until I make time to test things really well

cameronsutter (approved):

PR #458, opened 2020-11-14
Refactored main.js

shichongrui:


src/dashboard/index.js
```diff
@@ -6,10 +6,9 @@ import DashboardMain from './DashboardMain'
 const { remote } = require('electron')
 const app = remote.app
 
-import i18n from 'format-message'
-i18n.setup({
-  translations: require('../../locales'),
-  locale: app.getLocale() || 'en'
+i18n.setup({
```
cameronsutter, line 9: it looks like this file doesn't import i18n any more

bin/main_modules/menus/plottr.js
```diff
@@ -0,0 +1,147 @@
...
+      }
+    },
+    {
+      label: 'View Process Manager',
+      visible: NODE_ENV === 'dev',
+      click: () => openProcessManager()
+    },
+    {
+      label: i18n('Language'),
+      click: () => {
```
cameronsutter, line 96: is this where they are going to choose their language? I'm guessing this PR is just so I can see the refactor, but it's not ready to merge, right?

cameronsutter (commented):
Really liking how you organized things. I wasn't sure if moving some of those things to their own modules would be possible. Well done

bin/main_modules/menus/plottr.js
shichongrui: I'm going to do language selection in a different PR. I didn't mean for this to be included here, so I will remove it for this PR so this one can focus on the refactor.

src/dashboard/index.js
shichongrui: I'll fix it.

PR #459, opened 2020-11-17
Add migration for timelineScrollPosition

shichongrui:
refs: BAC-112


shichongrui:
@cameronsutter I'm not sure about how the date for the name of the migration is determined but I just set it to todays date.

cameronsutter (approved):
Date isn't that important as long as it's before the next release

PR #464, opened 2020-11-24
Switch to differential-update

shichongrui:
refs: BUG-49

According to https://github.com/electron-userland/electron-builder/issues/3106 the fix is to use `electron-differential-update` instead of `electron-update`


cameronsutter:
because of the conflict etc etc I just made the changes manually because it was so small

shichongrui:
@cameronsutter Were you able to test this on windows to make sure it actually works?

PR #465, opened 2020-11-24
Fix link rendering issue

shichongrui:
Also fixes a bug with pressing enter on the input to create the link in rich text

refs: BUG-71


src/app/components/rce/RichTextViewer.js
```diff
@@ -18,7 +18,7 @@ const RichTextViewer = (props) => {
 
   if (!value) return <span/>
   if (!value.length) return <span/>
-  if (value.length == 1 && value[0].children[0].text == '') return <span/>
+  if (value.length == 1 && value[0].children.length === 1 && value[0].children[0].text == '') return <span/>
```
shichongrui, line 21: There is obviously some history here and I tried not to remove any of it, but I added in an additional check since this was breaking the scenario of a single link.

cameronsutter (approved):

PR #469, opened 2020-12-01
Fix trial mode

shichongrui:
refs: BUG-87

Sorry for breaking this.


shichongrui:
Hmm it looks like `TRIALMODE` is set in quite a few places so I will need to change how setting the `TRIALMODE` works.

cameronsutter (approved):

PR #476, opened 2021-01-12
Bugfix drag and drop between rows in book list

Quiescent:
# BufFix: Dragging Between Rows in Book List

Fixes problem where books couldn't be dragged between rows of books in the BookList.

## Problem Description

When there are enough books, the display wraps books which don't fit on a single line onto the next.  `react-beautiful-dnd` is designed to handle a single row or column and fails to meet our need of dragging books in a flex-grid layout.

This problem is documented on Github: (https://github.com/atlassian/react-beautiful-dnd/issues/316).
There is mention of a plan to fix it, but, for the interim, a workaround is required.

## Alternatives

Other libraries exist which handle our use case.

### React Grid DnD

This is a good alternative and is worth considering as an alternative to this PR https://www.npmjs.com/package/react-grid-dnd.  I've used the library before and can confirm that it works for our use case.

### Use Multiple Droppables

As documented in `react-beautiful-dnd`'s docs, there's a way to transfer items between Droppables in the same DragAndDropContext.  (See https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/about/examples.md, https://codesandbox.io/s/-w5szl?file=/src/index.js).

This sample requires minor modifications and rebalancing between droppable areas to create the effect that there is one contiguous area.

### Others

Other developers report https://github.com/clauderic/dnd-kit and https://github.com/clauderic/react-sortable-hoc.

## Decision

I opted for Multiple Droppables because `react-beautiful-dnd` does what it says on the tin.  It's animations, features and look & feel are the better than React Grid DnD was when I used it.  It also appeared to be a simpler path than ripping out the existing work (I might be wrong here!)

### State

I used state on the BookList component because the fact that there are rows, the size of books and the number of books per row can be safely encapsulated without impacting the Redux store.  The purist side of me says that it should be in the store, but there should be no issues whatsoever reproducing the visual state of the UI from the store without including these pieces of state there.

### Resizing

I've hooked a method to update layout into the `resize` event so that Droppables are created dynamically and as needed when the window changes size.

When the component mounts, two refs are inspected to find out the size of books and the width of the DnD area.  This is potentially brittle because it relies on a DOM query to find the book with ID 0.

### Layout

When books are dragged downwards they leave behind a temporary gap in the source row.  This leads to a minor glitch from the users perspective when the layout updates because the first book in the destination row must move back up to the previous row.  The change only happens when the user drops the book so it might be a bit jolting for their experience.

## Testing

I tested the feature manually because I couldn't find many tests.  Is there a convention for tests or is this something that's open for design/discussion?

### DIFF
--- a/src/app/components/story/BookList.js
+++ b/src/app/components/story/BookList.js
@@ -60,10 +60,12 @@ class BookList extends Component {
     const sourceRow = +source.droppableId
     const destinationRow = +destination.droppableId
     const allIds = _.flatten(this.state.rows)
+    // Maintains relative positioning at destination of drop
+    const adjustForDownwardMovement = (sourceRow < destinationRow ? -1 : 0)
     const reOrderedIds = this.reorder(
       allIds,
       source.index + sourceRow * this.state.itemsPerRow,
-      destination.index + destinationRow * this.state.itemsPerRow
+      destination.index + destinationRow * this.state.itemsPerRow + adjustForDownwardMovement
     )
     this.setState({
       rows: _.chunk(reOrderedIds, this.state.itemsPerRow)
--- a/src/app/components/story/BookList.js
+++ b/src/app/components/story/BookList.js
@@ -29,6 +29,12 @@ class BookList extends Component {
   addBook = () => {
     const { actions, books, lineActions, sceneActions } = this.props
     const newBookId = objectId(books.allIds)
+    this.setState({
+      rows: [
+        ...this.state.rows.slice(0, this.state.rows.length - 1),
+        [...this.state.rows[this.state.rows.length - 1], newBookId]
+      ]
+    })
     actions.addBook()
     // add a plotline
     lineActions.addLine(newBookId)
--- a/src/app/components/story/BookList.js
+++ b/src/app/components/story/BookList.js
@@ -71,7 +71,7 @@ class BookList extends Component {
     this.props.actions.reorderBooks(reOrderedIds)
   }
 
-  componentDidMount () {
+  updateLayout = () => {
     let newBookWidth
     if (this.bookRef.current) {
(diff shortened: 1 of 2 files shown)


src/app/components/story/BookList.js
```diff
@@ -11,39 +11,114 @@ import Book from './Book'
 import { Glyphicon } from 'react-bootstrap'
 import { DragDropContext, Droppable, Draggable } from 'react-beautiful-dnd'
 import cx from 'classnames'
+import _ from 'lodash';
```
cameronsutter, line 14: i've started importing lodash like this:
`import { chunk } from 'lodash'`
I think it makes the bundle smaller. That's what I'm going for anyway

cameronsutter (approved):
I don't know if any of my comments are worth another commit, so i'm going to just approve

src/app/components/story/BookList.js
```diff
@@ -11,39 +11,114 @@ import Book from './Book'
...
 import { objectId } from '../../store/newIds'
 
 class BookList extends Component {
+  dragDropAreaRef = React.createRef();
```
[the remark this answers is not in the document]
Quiescent: Sorry!  I did actually pick up on semi-colons elsewhere and deleted them, but this stuff becomes muscle memory with time! XD

src/app/components/story/BookList.js
Quiescent: Noted.  Will do in the future.
(I actually searched the project to find the convention and found that it was being done this way.  I'll port any others that I find in the future).

Quiescent:
> 
> 
> I don't know if any of my comments are worth another commit, so i'm going to just approve

I went ahead and fixed the issues anyway.   Thanks for picking them up :)

cameronsutter (approved):

PR #478, opened 2021-01-18
Draft: Add prettier and eslint

Quiescent:
# Draft: Adds Prettier and EsLint Config

## Motivation
 - The team is growing in size, and it's easy to lose consistency in coding-style.
 - There are some trivial errors which slipped through the cracks.  EsLint picks them up before they reach clients.
 - I feel more confident about quality when I use EsLint.
 
## To Consider
Prettier is opinionated.  There are certain aspects of our style which it doesn't adhere to.  For instance: if you ask it add commas at the ends of lists then it also wants them at the ends of objects, without exception.

I've set it up to:
 - Demand no semicolons.
 - Error on unused variables.
 - Tab width of 2 spaces.
 - Check various React linting suggestions, such as:
    - Using deprecated APIs (which we do quite a bit, unfortunately).
    - Checking that all props are declared in the static prop types.
    - etc.
 - Check for single quote strings everywhere except in JSX tags' properties (those should be double quotes by convention).
 - Change layout for JS and JSX so that it's more readable (prettier has very concrete conventions here which are difficult to modify!)

IMO: We should definitely add more EsLint rules as we discover problems to prevent(!)


cameronsutter (approved):
another growing pain in the growth from side project to real product. Thanks for tackling this

PR #479, opened 2021-01-18
Add reloading

Quiescent:
# Adds Non-Hot Reloading
This pull request adds reloading when changes are made to the main process or render processes.
 - When a change is made to a renderer, then only that window is reloaded.
 - When a change is made to the main process, then the entire app is reloaded.

# Considerations
This approach uses a hard refresh.  It's better than nothing, but it's not HMR (https://webpack.js.org/guides/hot-module-replacement/).

# Improvements
After some additional research, I discovered that it's possible to electron to work with HMR, (which is lightning fast! :) ).  In general, there are two approaches:
 1. Use `electron-webpack`.
 2. Roll your own.

## Electron Webpack
Unfortunately, this approach would mean a major overhaul of our build system.  `electron-webpack` is a `cli` which bundles all of the `webpack` configurations.

https://webpack.electron.build/

## Roll Your Own
It's possible to roll your own, but it's quite an involved process.  I went a few steps in and discovered that we'd need to provide two versions of our HTML entry points ver every app and two separate webpack configuration files (to keep things a sane size!) if we went this route.

The idea is that you use ordinary HMR (with the webpack dev server) and then tell electron to load the bundle from a server instead of a static file while in development mode.

Here's an example of a project which does this: https://github.com/ccorcos/dev-tool


cameronsutter (approved):
looks good. You just need to fix the conflict in package-lock.json

Quiescent:
> 
> 
> looks good. You just need to fix the conflict in package-lock.json

Done.  I'll do the same for the third, but I suspect that I'll also have to rebase it once this is merged :)

cameronsutter:
i may have merged them in the wrong order. Now this one has conflicts again … sorry

Quiescent:
> 
> 
> i may have merged them in the wrong order. Now this one has conflicts again … sorry

No problem!  I'll push an update :)

PR #480, opened 2021-01-18
Add attributes to scenes (with template saving, but not usage)

Quiescent:
# Adds Attributes to Scene Cards
Just like characters can have attributes, so too can scene cards have attributes!
This task involved:
 - inspecting what had been done in terms of custom attributes already,
 - wiring up the existing reducers and actions,
 - refactoring out shared components, and
 - modifying the UI for editing scene cards to support attributes.

### Plot Line Control Bar
In the Scene view, there's now an attributes button which looks identical to that of the Characters view.  It opens a form for adding attributes which is shared with the characters view.  Sharing the form required some minor refactoring to pull out things like the entity which the properties come from.

### Card Dialog
I found that attributes don't fit well into the CardDialog component.  If you put them at the bottom then you either have to squash the main feature of editing a card (the scene description) or you have to overflow on Y.  I didn't think that the overflow looked great and the user might never notice that there are attributes at all, so I made the view tabbed and put the attributes on a second tab in the CardDialog.

An important note here: because of the way that forms work right now (see "Some Issues Encountered" for details) I had to make sure that the component is always mounted and the visibility changes because otherwise the child refs don't exist and we can't update the attributes (or we would clobber them with empty values).

# Some Issues Encountered

## Forms
The forms for attributes (and I'm not sure whether all forms work this way or not) use refs to select the form control and grab the value off of it.  Not being controlled in this way makes it more difficult to refactor them.  What I've done is collected child refs by passing a call back down into the child component and aggregating the refs in the parent.

It's far from ideal because we're now using state to track something that's non-functional.  I also noticed that the entire approach is deprecated because it doesn't work with function components(!)

I'd highly recommend that we investigate using `redux-forms` going forward.  They work really well and make some of the complicated logic we have around updating attributes much, much simpler.

I didn't change the way that we do forms in this pull request because it's my first task and I also think that it's a separate concern which warrants individual testing.

## Attribute Name Uniqueness
It used to be possible to add two custom attributes with the same name.  Ostensibly this is fine, but the app used the name as a unique id, so when you try to delete an attribute with that name then things got a bit wonky.  I've added a check for multiple attributes with the same name in the reducer.

We might need to consider the UX here.  I don't know how we notify the user of erroneous behaviour.

## State Arrays
I noticed that we have a tendency of storing objects in arrays in the redux store.  e.g. custom attributes for characters are an array.

This means that updates, additions etc are slow and error-prone.  The only reason I could think that we'd want them to be arrays is that we could have duplicates, but that seems to be a misfeature to me(?)

If we used objects instead then we'd get duplicate checking out fo the box, a big performance increase, and, with the object interpolation syntax they'd be easy to update too :)  e.g. the following would update or add a new entity to the store.

```javascript
return {
  ...state,
  [action.attributes.name]: action.attributes
}
```

# Not Addressed (Yet!)
You can create templates, but you can't create scene cards from templates yet.  I think that the control should be the same as adding a plot line using a template.  i.e. you hover over the tile to create a new one and if the cursor is to the left of the tile then it's added from a template.


cameronsutter:
Re: State Arrays

Yeah, that's something we should change. I've learned the error of my ways. To change things like this it requires a migration in the pltr repo. I'm not sure it makes sense to do that in this PR, but we should address that soon.

I've got a few other changes to the data structure that I'd like to make. Maybe it makes sense to do them all at the same time

Quiescent:
> 
> 
> Re: State Arrays
> 
> Yeah, that's something we should change. I've learned the error of my ways. To change things like this it requires a migration in the pltr repo. I'm not sure it makes sense to do that in this PR, but we should address that soon.
> 
> I've got a few other changes to the data structure that I'd like to make. Maybe it makes sense to do them all at the same time

Sounds good.  Perhaps it might be worthwhile to come up with a robust automatic way of migrating between versions and test it on something low-stakes.  Then we can make bigger migrations with more confidence.  WDYT?

cameronsutter:
> Perhaps it might be worthwhile to come up with a robust automatic way of migrating between versions and test it on something low-stakes. Then we can make bigger migrations with more confidence. WDYT?

Yeah, we should talk about that more

cameronsutter:
Re: Redux Forms

Never heard of that library, but it looks useful. The homepage says the creator has a "superior" library called React Final Form. Do you have a preference between the two?

Quiescent:
> 
> 
> Re: Redux Forms
> 
> Never heard of that library, but it looks useful. The homepage says the creator has a "superior" library called React Final Form. Do you have a preference between the two?

Hadn't actually heard of React Final Form.  I can check it out and let you know?

cameronsutter:
Re: Component always being mounted

I'm a little concerned about this. The timeline view has slow performance at times when there are many chapters and many plot lines so the less components mounted the better.

Is there any other way around mounting it always?

> otherwise the child refs don't exist and we can't update the attributes (or we would clobber them with empty values).

And could you point me to which child refs?

Quiescent:
> 
> 
> Re: Component always being mounted
> 
> I'm a little concerned about this. The timeline view has slow performance at times when there are many chapters and many plot lines so the less components mounted the better.
> 
> Is there any other way around mounting it always?
> 
> > otherwise the child refs don't exist and we can't update the attributes (or we would clobber them with empty values).
> 
> And could you point me to which child refs?

I might've given you the wrong idea.  What I mean to say is that while the CardDialog is open, both the Description RichText component and the list of custom attribute fields are mounted, despite the fact that only one is visible.

Once the user closes the CardDialog these components unmount as usual.
The penalty is the same as if we were showing all the input fields instead of using tabs on the CardDialog.

I'll comment on the regions of interest for this discussion.  One sec.

src/app/components/timeline/CardDialog.js
```diff
@@ -57,9 +70,62 @@ class CardDialog extends Component {
...
 
+  addInputRef = (ref) => {
+    // Set state needs to be queued so that React does not overwrite
+    // the attribute in a merge
+    setTimeout(() => {
+      this.setState({
+        inputRefs: [...this.state.inputRefs, ref],
+      })
+    }, 0)
+  }
```
Quiescent, line 81: This adds a ref to the state of the parent component.  The ref points at the input which edits the custom attribute.

src/app/components/timeline/CardDialog.js
```diff
@@ -57,9 +70,62 @@ class CardDialog extends Component {
...
+
+  findChildInput = (id) => {
+    const foundInput = this.state.inputRefs.find((ref) => {
+      return ref && ref.props.id === id
+    })
+    return foundInput && findDOMNode(foundInput)
+  }
```
Quiescent, line 88: This searches for a child ref which matches the name of the field we're looking for.

src/app/components/timeline/CardDialog.js
```diff
@@ -57,9 +70,62 @@ class CardDialog extends Component {
...
+    const attrs = {}
+    this.props.customAttributes.forEach(attr => {
+      const { name, type } = attr
+      if (type == 'paragraph') {
+        attrs[name] = this.state.paragraphs[name] || this.props.card[name]
+      } else {
+        const val = this.findChildInput(`${name}Input`).value
+        attrs[name] = val
+      }
+    })
```
Quiescent, line 110: This looks for values for all of the custom attributes.  If a child is unmounted then the ref isn't valid.
(BTW string-refs are also deprecated)

cameronsutter: re: string-refs

that'll be a refactor that touches just about everything, but we should stop using those

src/app/selectors/customAttributes.js
```diff
@@ -1,6 +1,7 @@
 import { createSelector } from 'reselect'
 import { allCharactersSelector } from './characters'
 import { allPlacesSelector } from './places'
+import { allCardsSelector } from './cards'
```
cameronsutter, line 4: was this change committed? I don't see it, but I might just be missing it

src/app/selectors/customAttributes.js
```diff
@@ -9,6 +10,7 @@ const placeKeys = Object.keys(place)
 export const allCustomAttributesSelector = state => state.customAttributes
 export const characterCustomAttributesSelector = state => state.customAttributes.characters
 export const placeCustomAttributesSelector = state => state.customAttributes.places
+export const scenesCustomAttributesSelector = state => state.customAttributes.scenes
```
cameronsutter, line 13: all the changes to this file will need to be mirrored to the pltr repo (and that change above from the selectors/cards.js file)

src/app/selectors/customAttributes.js
cameronsutter: or maybe that selector already existed?

cameronsutter: nvm

src/app/selectors/customAttributes.js
Quiescent: Ahh.  You mentioned that some of the logic hadn't been extracted, but it didn't quite click for me that there was duplication.
Let me see about migrating anything I've touched.

cameronsutter (approved):

PR #481, opened 2021-01-19
Add support for exporting characters, notes, and places

shichongrui:
Also fixes a bug with unicode conversion for rtf
Fixes a bug with headings not showing up in rtf documents
Fixed eslint and prettier conflict with spaces before function parens
Fixes scrivener exports for lists

fixes: KAN-12

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,6 @@ module.exports = {
   rules: {
     'no-unused-vars': ['warn', { args: 'none', argsIgnorePattern: 'req|res|next|val' }],
     'prettier/prettier': ['error'],
-    'space-before-function-paren': ['error', 'always'],
   },
   settings: {
     react: {
--- a/src/app/selectors/characters.js
+++ b/src/app/selectors/characters.js
@@ -1,6 +1,10 @@
 import { createSelector } from 'reselect'
 import { sortBy, groupBy } from 'lodash'
-import { characterSortSelector, characterFilterSelector } from './ui'
+import {
+  characterSortSelector,
+  characterFilterSelector,
+  currentTimelineSelector
+} from './ui'
 
 export const allCharactersSelector = state => state.characters
 // this one also lives in ./customAttributes.js but it causes a circular dependency to import it here
@@ -95,4 +99,14 @@ function sortEachCategory (visibleByCategory, sort) {
 export const charactersSortedAtoZSelector = createSelector(
   allCharactersSelector,
   (characters) => sortBy(characters, 'name')
+)
+
+export const charactersSortedInBookSelector = createSelector(
+  charactersSortedAtoZSelector,
+  currentTimelineSelector,
+  (characters, bookId) => characters.filter(character => {
+    if (character.bookIds.length === 0) return true
+    if (character.bookIds.includes('series')) return true
+    return character.bookIds.includes(bookId)
+  })
 )
--- a/src/common/exporter/scrivener/v2/exporter.js
+++ b/src/common/exporter/scrivener/v2/exporter.js
@@ -1,20 +1,17 @@
 import path from 'path'
 import fs from 'fs'
-import { cloneDeep, keyBy } from 'lodash'
 import log from 'electron-log'
-import { shell } from 'electron'
-import i18n from 'format-message'
 import xml from 'xml-js'
 import rtf from 'jsrtf'
-import { sortedChaptersByBookSelector, makeChapterTitleSelector } from '../../../../app/selectors/chapters'
-import { cardMapSelector } from '../../../../app/selectors/cards'
-import { sortCardsInChapter, cardMapping } from '../../../../app/helpers/cards'
-import { isSeriesSelector } from '../../../../app/selectors/ui'
-import { sortedLinesByBookSelector } from '../../../../app/selectors/lines'
 import serialize from '../../../slate_serializers/to_rtf'
 import { notifyUser } from '../../notifier'
+import exportChapters from './exporters/chapters'
+import exportCharacters from './exporters/characters'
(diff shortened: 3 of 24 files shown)


shichongrui:
@cameronsutter Requesting review from you because I want to make sure the format for the RTF docs makes sense for Scrivener.

.eslintrc.js
```diff
@@ -17,7 +17,6 @@ module.exports = {
   rules: {
     'no-unused-vars': ['warn', { args: 'none', argsIgnorePattern: 'req|res|next|val' }],
     'prettier/prettier': ['error'],
-    'space-before-function-paren': ['error', 'always'],
```
cameronsutter, line 20: i personally like this rule, though it's not a big deal either way, but we could just warn instead. Was it making something impossible?

src/app/selectors/characters.js
```diff
@@ -96,3 +100,13 @@ export const charactersSortedAtoZSelector = createSelector(
   allCharactersSelector,
   (characters) => sortBy(characters, 'name')
 )
+
+export const charactersSortedInBookSelector = createSelector(
```
cameronsutter, line 104: good idea. I wouldn't have thought of this. We just can't forget to mirror this to the pltr repo

src/common/exporter/scrivener/v2/exporter.js
```diff
@@ -1,31 +1,28 @@
...
 import xml from 'xml-js'
 import rtf from 'jsrtf'
-import { sortedChaptersByBookSelector, makeChapterTitleSelector } from '../../../../app/selectors/chapters'
-import { cardMapSelector } from '../../../../app/selectors/cards'
-import { sortCardsInChapter, cardMapping } from '../../../../app/helpers/cards'
-import { isSeriesSelector } from '../../../../app/selectors/ui'
-import { sortedLinesByBookSelector } from '../../../../app/selectors/lines'
 import serialize from '../../../slate_serializers/to_rtf'
 import { notifyUser } from '../../notifier'
+import exportChapters from './exporters/chapters'
```
cameronsutter, line 8: love this organization with splitting out these files. Good thinking

src/common/exporter/scrivener/v2/exporters/characters.js
```diff
@@ -0,0 +1,57 @@
...
+} from '../utils'
+
+export default function exportCharacters(state, documentContents) {
+  const { binderItem } = createFolderBinderItem(i18n('Characters'))
+  const characters = charactersSortedInBookSelector(state)
+  const allCharacterCategories = Object.values(characterCategoriesSelector(state))
+  characters.forEach((character) => {
+    const {
+      // these attributes are not being used
+      /* eslint-disable */
```
cameronsutter, line 19: this is the thing i really don't like about using code style tools

src/common/exporter/scrivener/v2/utils.js
```diff
@@ -0,0 +1,171 @@
...
+export function isPropertyEmpty(property) {
+  if (!property) return true
+
+  // Rich content is an array
+  const isRichContent = Array.isArray(property)
+
+  // If it isn't rich content and it's not empty then it has a value
+  if (!isRichContent) return false
+
+  // if it's empty it only has one item and the type is blank
```
cameronsutter, line 151: this one might not actually be accurate. This is the default/initial value for rich content:

`[{ type: 'paragraph', children: [{ text: '' }] }]`

Before, the initial value didn't have a `type: 'paragraph'` so I wonder if the file you're working with is like that maybe?

src/common/slate_serializers/to_rtf.js
```diff
@@ -91,15 +99,15 @@ const leafInElement = (node) => {
...
+// There seems to be one thing about these lists that is weird
+// The first list item has some extra spacing between the left margin
+// and the bullet point. The rest of the items do not have this and I
+// can't figure out why it is there. (Matt)
 const makeListItem = (node) => {
   const text = node.children.flatMap(ch => ch.text).join(' ')
-  return rtf.Utils.makeRtfCmd("{\\listtext	\\uc0\\u8226	}", text, "\\\n", false)
-}
\ No newline at end of file
+  return `{\\pard\\fi-300\\li100\\bullet\\tab ${text} \\par}`
```
cameronsutter, line 112: is there some documentation that you found for rtf documents? It might be valuable to keep that somewhere we can all find (maybe even just a link in a comment)

cameronsutter (approved):
The only comments you need to possibly make a change for, are 1) the question in .eslintrc.js and 2) the comment about default/initial values for rich content in exporter/scrivener/v2/utils.js

If you do end up changing them, feel free to merge the PR once you do, unless you want me to take a look again

.eslintrc.js
Quiescent: There's actually a problem with this rule which Matthew might have run into.  `eslint` will let you configure it to ask for these spaces, but I discovered yesterday that `prettier` doesn't listen to `eslint` and will always insist that there's no space.  Unfortunately, there's no option to configure `prettier` to ask for spaces.  A fork of `prettier` exists which allows for these spaces, but it looks so unpopular that I'm not sure that it's a good idea to go for it.

shichongrui: Yep it was because of the conflict between what the two linters wanted.

src/common/exporter/scrivener/v2/utils.js
shichongrui: Ok I will add this as well.

src/common/slate_serializers/to_rtf.js
shichongrui: Yeah I can add it to this file.

PR #484, opened 2021-01-23
Enable templates for scene creation

Quiescent:
# Motivation
To enable users to create scene cards from custom attribute templates.

# UI Modifications
This branch adds UI elements for creating a new card with a template and picking a scene template.

## Creating a Scene with a Template
I copied the way that plotlines are created with templates so that:
 - when you hover of a blank card you see a split button (like plotlines);
 - when you click on the plus icon, we enter the original flow for creating a scene card (i.e. title form appears).
 - when you click on "Use Template" it opens the template picker and when you've picked a template, it advances to the title input form and remembers the template selection.

## Picking a Scene Template
We now have a `SceneTemplateDetails` component wired up to the Template picker when scene templates are selected.

# Non-Functional Changes
The following changes were made to the plumbing to support the UI changes:
 - Reducers and actions for `lines` were migrated to `pltr`.
 - Lines now have a `backgroundColor` to keep the new scene buttons consistent with the line colours.
 - `EditAttribute` now handles template attributes.
 - Blank project creation now includes background colours for lines.

### DIFF
--- a/shared/initialState.js
+++ b/shared/initialState.js
@@ -141,6 +141,7 @@ const line = {
   id: 1,
   bookId: 1,
   color: '#6cace4',
+  backgroundColor: '#6cace419',
   title: '',
   position: 0,
   characterId: null,
@@ -151,6 +152,7 @@ const line = {
 const seriesLine = {
   id: 1,
   color: '#6cace4',
+  backgroundColor: '#6cace419',
   title: '',
   position: 0,
   expanded: null,
--- a/src/app/components/timeline/BlankCard.js
+++ b/src/app/components/timeline/BlankCard.js
@@ -6,9 +6,10 @@ import { bindActionCreators } from 'redux'
 import { Cell } from 'react-sticky-table'
 import * as CardActions from 'actions/cards'
 import i18n from 'format-message'
-import { FormControl, FormGroup, ControlLabel } from 'react-bootstrap'
+import { FormControl, FormGroup, ControlLabel, Glyphicon } from 'react-bootstrap'
 import cx from 'classnames'
 import { isSeriesSelector } from '../../selectors/ui'
+import TemplatePicker from '../../../common/components/templates/TemplatePicker'
 
 class BlankCard extends Component {
   constructor(props) {
@@ -16,6 +17,10 @@ class BlankCard extends Component {
     this.state = {
       creating: false,
       dropping: false,
+      templateHover: false,
+      defaultHover: false,
+      showTemplatePicker: false,
+      templates: [],
     }
   }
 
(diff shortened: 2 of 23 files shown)


shared/initialState.js
```diff
@@ -140,6 +140,7 @@ const line = {
   id: 1,
   bookId: 1,
   color: '#6cace4',
+  backgroundColor: '#6cace419',
```
cameronsutter, line 143: oh, i see now what you meant about this. Why does this have to be a separate attribute?

src/app/components/timeline/BlankCard.js
```diff
@@ -6,119 +6,225 @@ import { bindActionCreators } from 'redux'
...
-    var blankCardStyle = {
-      borderColor: this.props.color
+    const blankCardStyle = {
+      borderColor: this.props.color,
+      color: this.props.color,
     }
-    return <div className={cx('blank-card__body', {hover: this.state.dropping})} style={blankCardStyle} />
+    const addWithTemplateStyle = this.state.templateHover
+      ? {
+          background: this.props.backgroundColor,
```
cameronsutter, line 161: couldn't this just be `this.props.color`?

cameronsutter (commented):
I know this is going to cause a lot of conflicts with my changes … yippeee

I think we should hold off merging this until next week (we're planning to release this week). Is that going to make any future changes for you more difficult?

shared/initialState.js
Quiescent: > I know this is going to cause a lot of conflicts with my changes … yippeee
> 
> I think we should hold off merging this until next week (we're planning to release this week). Is that going to make any future changes for you more difficult?

Nope.  I can totally hold off and keep rebasing my latest branches.  No problem :+1:

Quiescent:
> I know this is going to cause a lot of conflicts with my changes … yippeee
> 
> I think we should hold off merging this until next week (we're planning to release this week). Is that going to make any future changes for you more difficult?

Nope. I can totally hold off and keep rebasing my latest branches. No problem :+1:

shared/initialState.js
Quiescent: It gets added programmatically when we render the line, so I'm not sure how else to specify an exact colour.
So in this PR I only made it a less opaque version of the original colour, but I'm not sure that that's perfect, style-wise.
If we settle on it just being lighter, then we can compute it from color using something like `tinycolor` https://www.npmjs.com/package/tinycolor2.

I might be drawing blanks here though.  Can you think of something better?

src/app/components/timeline/BlankCard.js
Quiescent: It doesn't look good just using colour for the background colour.
It's way too strong for the text to come through and stands out too much on the UI.

PR #487, opened 2021-01-27
Release Bug Fix: blank cards error on close

Quiescent:
# Release Bug Fix: Blank Cards Error on Close
## Steps to Reproduce
Create a new blank card and chapter by clicking on the plus icon next to a card.
Open the card for editing.
Close the card.

Result: an exception is thrown.

## Proposed Fix
The feature to add a blank card when inserting in between cards creates cards without template values.
The fix (use an empty array if templates aren't present) is a bandage over this and similar causes of the problem because we aim to remove the feature of adding a blank card soon anyway.

### DIFF
--- a/src/app/components/timeline/CardDialog.js
+++ b/src/app/components/timeline/CardDialog.js
@@ -24,7 +24,7 @@ import { truncateTitle } from '../../helpers/cards'
 import { EditAttribute } from '../EditAttribute'
 
 class CardDialog extends Component {
-  constructor (props) {
+  constructor(props) {
     super(props)
     this.state = {
       description: props.card.description,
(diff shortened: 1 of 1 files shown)


src/app/components/timeline/CardDialog.js
```diff
@@ -367,7 +384,9 @@ class CardDialog extends Component {
                 </div>
               </div>
               <div
-                className={cx('card-dialog__details-show-hide-wrapper', {hidden: selected !== 'Description'})}
+                className={cx('card-dialog__details-show-hide-wrapper', {
```
cameronsutter, line 387: is there any way we can make the linter/prettier keep these on one line? It just feels like in this case it makes it harder to read instead of nicer.

cameronsutter (approved):

PR #488, opened 2021-01-27
Filter scenes by custom attributes

Quiescent:
# Add the Ability to Filter on Custom Attributes
This increases the utility of custom attributes on the timeline view.
Fields to filter on are automatically generated from the values of custom attributes on the scene cards.

## Implementation Details
The TimeLineWrapper now uses `CustomAttrFilterList`, like the `CharacterListView`.
It was generalised to support different built-in attributes to filter on, and a few switch statements to determine which selectors etc. to use.

## Other Areas Impacted
The following areas of the app also use the `CustomAttrFilterList`:
 - `PlaceListView`;
 - `CharacterListView`;

Both were tested, and no faults were found.

# Improvements
There are some places which define which attributes should be filtered.
The definition looks something like this: `{ books: [], places: [] ...}`.
I don't think that it's necessary to define these objects at all now that we've started adding fields to the filter object when they're missing.
The selectors define the contract so we could remove this redundancy.

# Considerations
Template attributes live on a different part of the object so that they don't class with manual attributes.
This branch doesn't address that concern yet because templates are under development.


cameronsutter:
Re: Improvements

@Quiescent Do you mean that you've made these improvements or that they are ideas for future improvements?

Quiescent:
> Re: Improvements
> 
> @Quiescent Do you mean that you've made these improvements or that they are ideas for future improvements?

Future improvements.  These thoughts occurred to me when I authored the PR, and I thought that I'd rather note them here than forget them :)

cameronsutter (approved):

PR #489, opened 2021-01-27
Don't save attributes on scenes when cancelling the dialog

Quiescent:
# Bug Fix: Don't Save Attributes When Closing Scene Card Dialogs
This removes the call to save in `componentWillUnmount`.
It has the knock-on effect of not double saving when we hit `save`.

# Considerations
Hitting `esc` still saves and closes.
Is that desirable?

# Updates
After @cameronsutter 's comments, the implementation changed.
It now uses a piece of state to flag that we're unmounting and deliberately preventing a save.
i.e. the default behaviour is to save when the component unmounts.
It will only *not* save when asked to.

### DIFF
--- a/src/app/components/timeline/CardDialog.js
+++ b/src/app/components/timeline/CardDialog.js
@@ -97,7 +97,8 @@ class CardDialog extends Component {
   }
 
   saveAndClose = () => {
-    this.saveEdit()
+    // componentWillUnmount saves the data (otherwise we get a
+    // duplicate event).
     this.props.closeDialog()
   }
 
@@ -194,6 +195,7 @@ class CardDialog extends Component {
     this.setState({
       addingAttribute: true,
     })
+  }
 
   closeWithoutSaving = () => {
     this.setState(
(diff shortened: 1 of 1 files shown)


cameronsutter:
one of the problems with doing it this way is that the save on `componentWillUnmount` was to prevent people from losing data when they switched tabs. So let's say someone is typing the card dialog, and they click the characters tab before closing this card 💥  everything they wrote is gone.

Maybe there is a state value that says the user is attempting to cancel and when that is true, don't save. This is a little janky but I've had to do this elsewhere i think

Quiescent:
> one of the problems with doing it this way is that the save on `componentWillUnmount` was to prevent people from losing data when they switched tabs. So let's say someone is typing the card dialog, and they click the characters tab before closing this card boom everything they wrote is gone.
> 
> Maybe there is a state value that says the user is attempting to cancel and when that is true, don't save. This is a little janky but I've had to do this elsewhere i think

Noted.  One moment.

Quiescent:
@cameronsutter how about this solution?

Note: I also prevented the double save in this one.

PR #494, opened 2021-02-01
Mass migration to pltr 2

Quiescent:
# Moves All Reducers, Actions, Selectors etc. to pltr
Also embeds pltr into plottr_electron.

### Migration
The migration was made by diffing each file and ensuring that pltr had at least as much functionality as plottr_electron.
A downside of this approach is that pltr might have some redundant functions left in it.

### Embedding Pltr
Pltr was embedded as a subtree.
The following command created the subtree:
```
git subtree add --prefix lib/pltr pltr master --squash
```

Here are some useful commands to use the subtree:
```
git remote add -f pltr git@github.com:Plotinator/pltr.git
git subtree pull --prefix lib/pltr pltr master --squash
git subtree push --prefix=lib/pltr/ pltr <your-branch-name-here>
```


cameronsutter:
Can you add those useful commands to the readme file?

PR #503, opened 2021-02-08
Merge chapters into beats

Quiescent:
# Merge Chapters Into Beats
Conceptually, chapters and beats are very similar concepts.
Both model the idea of literary content which is organised into shorter sections with overarching themes.
The decision was taken to make a chapter a specialisation of a beat because:
 - the actions, reducers, selectors etc. for chapters and beats are very similar,
 - beats model the broader scoped-idea of themes across titles,
 - fewer flag checks are required throughout the app (rendering a beat or a chapter is similar enough that no check is actually required in almost all cases), and
 - new features which augment chapters were not present for beats (and now are because they're the same concept).

## Important Consideration
How do we ensure that changes to the Redux store structure are compatible with templates?
Have we already solved this problem?

## Individual Changes
This branch makes various miscellaneous changes to the project:
 - renaming chapter to beat in all properties,
 - deleting chapter selectors, reducers, constants, etc.
 - renaming chapter helpers,
 - renaming all components with the word chapter to be instead named `Beat<X>`,
 - renaming all `css` classes which mentioned chapters to be instead named `...-beat-...`
 - applies stricter rules to EsLint to help prevent re-naming errors by detecting unused and undefined variables (exclusions for global variables are defined in `.eslintrc.js`).
 - applies stricter rules to EsLint to help prevent errors in case blocks (detects lexically scoped variables which escape the current case, and what could be unintentional fall-through in case blocks).
 - various fixes to files across the project where variables were used without defining them, imports were missing, lexical variables escaped case blocks, unintentional fall-throughs, variable re-declarations because lexical keywords weren't used, and object keys were duplicated.

### 2021-2-8 Migration
The 08-02-2021 migration moves all chapters into beats, deletes chapters, gives the newly migrated beats fresh ids, and ensures that all cards point at the correct beats.

### Scene Actions
Scene actions don't exist anymore.
It appears that the word "Scene" was at some point used interchangeably with the word "Chapter".

### Scene Custom Attributes
Custom attributes for cards are called scene attributes.
This name was given in error.
These should, in fact, be called "card attributes."
Action types and function names have changed, but this PR doesn't change the name of these attributes in the Redux store, nor does it include a migration to change the structure of custom scene (card) attributes.

## Other Features Directly Impacted
The exporter/importers are all affected by this change.


cameronsutter:
Wow, what a PR. Well done

> It appears that the word "Scene" was at some point used interchangeably with the word "Chapter".

Yes, chapters used to be called scenes originally and there's still some remnants of that.

> this PR doesn't change the name of these attributes in the Redux store, nor does it include a migration to change the structure of custom scene (card) attributes.

that's fine

> How do we ensure that changes to the Redux store structure are compatible with templates?
Have we already solved this problem?

We haven't solved this yet.

src/app/components/timeline/BeatsCell.js
```diff
@@ -24,39 +24,39 @@ const {
   isMediumSelector,
 } = selectors
 
-class ScenesCell extends PureComponent {
+class BeatsCell extends PureComponent {
```
cameronsutter, line 27: This one actually should stay ScenesCell (or something like that - CardsCell?) because it's referring to a scene card

src/common/importer/snowflake/importer.js
```diff
@@ -267,14 +267,14 @@ function cards(currentState, json, bookId) {
       { position: 0, title: i18n('Main Plot') },
       bookId
     )
-    const scenesByChapter = groupBy(sceneListNode['GContainer'], (scene) => {
-      const node = scene['GInteger'].find((n) => n['_attributes']['name'] == 'chapter')
+    const scenesByBeat = groupBy(sceneListNode['GContainer'], (scene) => {
+      const node = scene['GInteger'].find((n) => n['_attributes']['name'] == 'beat')
```
cameronsutter, line 271: this 'chapter' comes from the Snowflake file, so it needs to stay 'chapter'

src/dashboard/components/files/FileActions.js
```diff
@@ -36,6 +36,8 @@ export default function FileActions({ missing, id, filePath, openFile }) {
     switch (eventKey) {
       case 'open':
         openFile(filePath, id)
+        shell.showItemInFolder(filePath)
```
cameronsutter, line 39: I think this line was added in error. It's not needed here, but the `break` below this line is definitely needed. Good catch

src/app/components/timeline/BeatsCell.js
Quiescent: I changed `BeatCardAdd` and `BeatsCell` to `CardAdd` and `CardCell` respectively.
I like the consistency that this has to `cards` in the store.
If we used `scene`, the reader might be misled into believing that another concept is encoded in the name.

src/common/importer/snowflake/importer.js
Quiescent: Oops!
I did look through the importer and exporter and missed that.

src/dashboard/components/files/FileActions.js
Quiescent: Ja.  I thought it might be but wasn't sure.
EsLint pointed it out as a potential mistake, and I wanted to start by retaining the behaviour while removing the fall-through.

I've gone ahead and removed it.

PR #508, opened 2021-02-11
Add an example for scrivener export and update docx export example

Quiescent:
# Improve the Testing Process Re. Exports
The process for verifying an exported file is tedious and time-consuming.
This PR intends to improve it so that we can make modifications faster and with greater confidence.
It does the following:
 - adds an integration test for the word exporter,
 - updates the example exported version of `example3.pltr`,
 - adds a Scrivener export example.

## Add an Integration Test for Word Exporter
The test does the following:
 - copies the current word export of `example3.pltr` to a `.prev` directory,
 - opens `example3.pltr`,
 - exports it to `example3.docx`,
 - extracts both the old and new `docx` documents into temp directories in the root of the repo,
 - parses the `word/document.xml` file (the most important file in a `docx` ITO contents) for both old and new into JSON,
 - asserts that they are the same (this will fail with intentional changes and produce the relevant diffs),
 - delete the temp directories.

## Update the Export Examples from `example3.pltr` for `docx` and `scrivener`
The example files and exported versions are made up to date by this PR.


package.json
```diff
@@ -140,7 +141,8 @@
     "uuid": "^8.3.1",
     "webpack": "^4.35.0",
     "webpack-bundle-analyzer": "^3.7.0",
-    "webpack-cli": "^3.3.5"
+    "webpack-cli": "^3.3.5",
+    "xml2js": "^0.4.23"
```
cameronsutter, line 145: we've got `xml-js` already in the project. Does this one do something different? Or is it better?

cameronsutter (commented):
Besides my comment on the `xml2js` package (which compared to the rest of the PR is not that big of a deal), i'm all for this change

package.json
Quiescent: Oh!  I didn't spot it.
One moment, please.

Quiescent: Done.

PR #514, opened 2021-02-14
Migrate templates before merging

Quiescent:
# Handle Old/Outdated Templates in Current Version
Problem: the user might create a template at some point in time and wish to use that template after upgrading Plottr.
Templates mimic the Plottr file, and new files merge these properties into the new file format.

## Solution
In this PR, we handle this situation by creating a temporary file from the template and then migrating it.

### Rationale
Migrations are the source of truth for how older formats should be treated.
We don't need to do anything risky with our user's templates with this approach, nor do we need to regularly update our own templates.
They should both just work :tm: !

The following adjustments were made to the create-from-template process:

## Put just enough into a template to make the migrator see it as a template
This process requires that the template have a few basic properties added to it so that the migrator can handle it (e.g. the property `file.version`.

## Migrations can't assume that any property is present on the file
All migrations now need to check that a property is present before modifying it or depending on it.
Otherwise, the migrator will bomb out when it migrates a template which doesn't contain a property which a migrator tries to migrate.
e.g.
```patch
modified   lib/pltr/v2/migrator/migrations/0.6.js
@@ -9,19 +9,25 @@ export default function migrate(data) {
   obj.file.version = '0.6.0'
 
   // add tag colors
-  obj.tags.forEach((t) => {
-    t['color'] = t['color'] || null
-  })
+  if (obj.tags) {
+    obj.tags.forEach((t) => {
+      t['color'] = t['color'] || null
+    })
+  }
```

# Attention
I'd like to draw the reviewers attention to the following block:
```javascript
const mergeRecursivelyWithArrayConcatStrategy = (object1, object2) => {
  const concatenateArrays = (objValue, srcValue) => {
    if (isArray(objValue)) {
      return objValue.concat(srcValue)
    }
  }
  return mergeWith(object1, object2, concatenateArrays)
}
```

The migrator for templates uses this method to merge into a new file.
This method concatenates arrays when it finds them rather than overriding attributes.
Is this the right approach?


cameronsutter:
re: mergeRecursivelyWithArrayConcatStrategy

That's feels like the right strategy, but not in all situations. For example with the plotlines (`lines`). The empty file will have one plotline, but when we merge it with a template, we don't want that original plotline to remain. We want the template to overwrite that.

The same holds true for beats, customAttributes, and maybe others that I can't remember right now. Now that i think about it, I can't think of any example where we wouldn't want the template to overwrite the empty file, but there might be an instance that I can't remember.

src/dashboard/components/files/FilesHome.js
```diff
@@ -16,7 +16,10 @@ export default function FilesHome(props) {
   const createWithTemplate = (template) => {
     MPQ.push('btn_create_with_template')
     try {
-      createNew(template.templateData)
+      createNew({
+        version: template.version,
```
cameronsutter, line 20: in the case of user templates, the version is a timestamp instead of a version number (e.g. 2021.2.4) so I think we'll need to convert this to a version number first here

cameronsutter (commented):
we use the templates in multiple places, which makes the JIT migrating less elegant because we’ll have to make sure it happens in each spot.

So the question is, do we want to 1) migrate templates all at once and save them to disk, or 2) use your idea of JIT migrating and make sure we get all the places where they are used?

As far as the built-in templates (in the pltr-templates repo) I think we should have a script I can run to migrate them every time there is a new migration to run

Quiescent:
> re: mergeRecursivelyWithArrayConcatStrategy
> 
> That's feels like the right strategy, but not in all situations. For example with the plotlines (`lines`). The empty file will have one plotline, but when we merge it with a template, we don't want that original plotline to remain. We want the template to overwrite that.
> 
> The same holds true for beats, customAttributes, and maybe others that I can't remember right now. Now that i think about it, I can't think of any example where we wouldn't want the template to overwrite the empty file, but there might be an instance that I can't remember.

Noted.  I'll go for the overwrite strategy then.
We might run into trouble in the future if we have nested structures, but for the time, I see what you're saying and agree.

Quiescent:
> we use the templates in multiple places, which makes the JIT migrating less elegant because we’ll have to make sure it happens in each spot.
> 
> So the question is, do we want to 1) migrate templates all at once and save them to disk, or 2) use your idea of JIT migrating and make sure we get all the places where they are used?
> 
> As far as the built-in templates (in the pltr-templates repo) I think we should have a script I can run to migrate them every time there is a new migration to run

I've replied on Slack.  I think that JIT is probably still best, and that we need a formalised API for migrating templates.
I'll get to work on a bulk-migration solution for the `pltr-templates` repo as well.

src/dashboard/components/files/FilesHome.js
Quiescent: I actually think that it's pretty good as is.
The timestamp puts the template in a time interval between two versions, so it's valid for the most recent older version of Plottr.
The migrator will treat this as a version and apply only migrations after that date.
This should also work for the older non-timestamp versions of Plottr.

Am I missing something?

cameronsutter (approved):

PR #518, opened 2021-02-18
Update trial_manager.js

murfdog64:
Forgot to add comments on my commit.. You'll see in the end I just kept one change to use the start date instead of the end date since we were just trying to fix the extension. I feel pretty confident in that much at least

### DIFF
--- a/src/common/licensing/trial_manager.js
+++ b/src/common/licensing/trial_manager.js
@@ -15,7 +15,7 @@ export function useTrialStatus() {
   }
 
   const extendTrial = (days) => {
-    const newEnd = addDays(trialInfo.endsAt, days)
+    const newEnd = addDays(trialInfo.startsAt, days)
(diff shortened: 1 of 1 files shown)


src/common/licensing/trial_manager.js
```diff
@@ -15,7 +15,7 @@ export function useTrialStatus() {
   }
 
   const extendTrial = (days) => {
-    const newEnd = addDays(trialInfo.endsAt, days)
+    const newEnd = addDays(trialInfo.startsAt, days)
```
cameronsutter, line 18: You found the right line to change. That's sometimes the hardest part.

Instead of sending `trialInfo.startsAt` to `addDays`, we want it to add 30 days to today's date. So you want `Date.now()` instead

cameronsutter (changes requested):

PR #519, opened 2021-02-23
Bug184

murfdog64:


src/common/licensing/trial_manager.js
```diff
@@ -15,7 +15,7 @@ export function useTrialStatus() {
   }
 
   const extendTrial = (days) => {
-    const newEnd = addDays(trialInfo.endsAt, days)
+    const newEnd = addDays(Date.now(), days)
```
cameronsutter, line 18: Weird that this showed up in this PR. Makes me think your branch isn't up to date with master

lib/pltr/v2/reducers/lines.js
```diff
@@ -43,7 +43,7 @@ export default function lines(state = initialState, action) {
         {
           id: nextId(state),
           bookId: actionBookId,
-          title: '',
+          title: i18n('Main Plot'),
```
cameronsutter, line 46: I don't think we want to change the reducer to always add this title.

Look in `project/BookList` where books get added. I think the solution lies there. And that might be what you meant when you were asking about `addLineWithValues`. I think you'll need to use that action

cameronsutter (changes requested):

src/common/licensing/trial_manager.js
murfdog64: This is showing up now.. branch is hopefully up to date now

cameronsutter (approved):

PR #521, opened 2021-03-03
Bug210

murfdog64:


src/app/components/export/ExportNavItem.js
```diff
@@ -20,7 +20,11 @@ function ExportNavItem(props) {
 
   const doExport = (type) => {
     let label = i18n('Where would you like to save the export?')
-    const defaultPath = path.basename(props.fileName).replace('.pltr', '')
+    const defaultPath =
+      props.bookId !== 'series'
+        ? props.books[props.bookId].title
+        : props.seriesName + ' (Series View)'
```
cameronsutter, line 26: you'll need to wrap `'(Series View)'` in a call to `i18n` like this: 

```
props.seriesName + ' ' + i18n('(Series View)')
```

src/app/components/export/ExportNavItem.js
```diff
@@ -20,7 +20,11 @@ function ExportNavItem(props) {
 
   const doExport = (type) => {
     let label = i18n('Where would you like to save the export?')
-    const defaultPath = path.basename(props.fileName).replace('.pltr', '')
+    const defaultPath =
+      props.bookId !== 'series'
+        ? props.books[props.bookId].title
```
cameronsutter, line 25: this is a quirk I need to fix, but `bookId` is a number whereas you need a string for `props.books[<id>]`, so for now you'll have to do:

```
props.books[`${props.bookId}`]
```

src/common/exporter/word/exporter.js
```diff
@@ -59,7 +59,9 @@ function namesMapping(data) {
 }
 
 function seriesNameSection(data, bookId) {
-  let titleText = isSeries(bookId) ? data.series.name : data.books[bookId].title
+  let titleText = isSeries(bookId)
+    ? data.series.name + i18n(' (Series View)')
```
cameronsutter, line 63: unfortunately i think we need the space outside of the translation

```
data.series.name + ' ' + i18n('(Series View)')
```

package.json
```diff
@@ -132,6 +132,7 @@
     "node-sass": "^4.14.1",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.2.1",
+    "prettier-eslint": "^12.0.0",
```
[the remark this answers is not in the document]
murfdog64: I'm not.. sure.. I installed it the same way I installed the other two.. I didn't notice anything using it specifically. But I guess it's hard for me to say since that was my first rodeo with it.

PR #524, opened 2021-03-09
adjustments for collapse/expand timeline on title cell and nav

murfdog64:


src/app/components/timeline/LineTitleCell.js
```diff
@@ -394,7 +419,7 @@ function mapStateToProps(state, ownProps) {
 
 function mapDispatchToProps(dispatch, ownProps) {
   return {
-    actions: bindActionCreators(LineActions, dispatch),
+    actions: bindActionCreators(Object.assign({}, LineActions, uiActions), dispatch),
```
cameronsutter, line 422: What we normally do here is add another line below this, like so:

```
uiActions: bindActionCreators(uiActions, dispatch),
```

src/app/components/timeline/LineTitleCell.js
```diff
@@ -219,9 +234,14 @@ class LineTitleCell extends PureComponent {
             <Glyphicon glyph="tint" />
           </Button>
           {isSmall ? null : (
-            <Button block bsSize="small" onClick={this.toggleLine}>
-              {expandedIcon}
-            </Button>
+            <ButtonGroup>
```
cameronsutter, line 237: does this look good as a `ButtonGroup`? I think it would look the same as two buttons next to each other, right?

cameronsutter (commented):
Just one small adjustment and it's good to go

src/app/components/timeline/LineTitleCell.js
murfdog64: done!

src/app/components/timeline/LineTitleCell.js
murfdog64: I put it them in the group to keep them within a parent element, but on closer inspection, in vertical it gets that -1px margin that throws it. So I should have just asked what type of parent you recommend.

cameronsutter (approved):

PR #528, opened 2021-03-11
Extract locales

Quiescent:
# Use Locales from `plottr_locales`
This PR makes `plottr_electron` dependent on `plottr_locales`.
It also uses the version of `format-message`, which `plottr_locales` exports.
The second point prevents package version conflicts (I ran into one when I initially created the dependency.)


cameronsutter:
I'm afraid i've made a few headaches for you.

1. If you merge master into this branch you'll notice the main process modules have been moved, so that might cause a lot of conflicts
2. Last year I started moving `i18n` to `t` (for translate), but didn't change it everywhere. `t` is a lot easier to type, so I was thinking it would be nicer to use it like that, but it seems to have caused problems everywhere in this PR. Thoughts on the Right Way™ to proceed?

Otherwise I like where you're going

Quiescent:
> 
> 
> I'm afraid i've made a few headaches for you.
> 
>     1. If you merge master into this branch you'll notice the main process modules have been moved, so that might cause a lot of conflicts
> 
>     2. Last year I started moving `i18n` to `t` (for translate), but didn't change it everywhere. `t` is a lot easier to type, so I was thinking it would be nicer to use it like that, but it seems to have caused problems everywhere in this PR. Thoughts on the Right Way™ to proceed?
> 
> 
> Otherwise I like where you're going

Great.  I don't think that it'll be a problem to change to `t` everywhere.
Update in coming!

Quiescent:
@cameronsutter new version is in, WDYT?

cameronsutter (approved):

PR #532, opened 2021-03-15
Bug207

murfdog64:


src/css/error_boundary.scss
```diff
@@ -42,4 +42,10 @@
   flex-direction: column;
   align-items: center;
   max-width: 60%;
-}
\ No newline at end of file
+}
+
+body.darkmode {
+  .error-boundary-title {
+    color: $blue-0;
```
cameronsutter, line 49: there's actually a `$darkmode-text` variable (or something like that)

package.json
```diff
@@ -73,7 +73,7 @@
     "css-loader": "^3.0.0",
     "devtron": "^1.2.1",
     "docx": "^5.0.2",
-    "electron": "^10.2.0",
+    "electron": "^10.4.0",
```
cameronsutter, line 76: this and the package-lock changes worry me. Maybe you can make a new branch from master and make just the two changes below.

cameronsutter:
we made another PR

PR #539, opened 2021-03-18
Use cross-env to simplify npm scripts on Windows

Quiescent:
# Motivation
Setting environment variables differs slightly between Windows and *Nix systems.
Instead of duplicating our scripts, we could use a nifty package like `cross-env` to handle that stuff for us(!)
This PR installs `cross-env` as a dev dependency and then uses it to simplify all of our npm scripts that set environment variables.


cameronsutter (approved):
Sweeeet. Good idea

cameronsutter:
Let's hold off on merging this until the release tomorrow. From now on we need to use a release branch to avoid this. Sorry

PR #544, opened 2021-03-22
kan-153 remove console errors from FormControl with inputRef

jeanawhou:


Quiescent (changes requested):
@jeanawhou looks great!
There's one small problem: if we're going to set all of these refs by callback then we shouldn't also call `React.createRef()` to initialise them in the constructor.  We should just set them to null there.

jeanawhou:
@Quiescent thank you, i have update the PR just now

Quiescent (approved):
Sorry for the delayed response.  Looks good! 👍

PR #545, opened 2021-03-23
Kan167

murfdog64:


src/dashboard/components/options/OptionsHome.js
```diff
@@ -9,10 +9,48 @@ import DarkOptionsSelect from './DarkOptionsSelect'
...
+  const renderDropdown = () => {
+    return (
+      <select onChange={(event) => setBackupType(event.target.value)}>
+        <option value="days">{t('Days of Backups')}</option>
+        <option value="number">{t('Number of Backups')}</option>
+      </select>
+    )
+  }
+  const displayBackupOption = () => {
+    switch (backupType) {
```
jeanawhou: I think the switch here is unnecessary, maybe we can use a ternary operator since the 2 <FormControl /> are almost the same

Something like:

```suggestion
      <div className="backup-type">
                <ControlLabel>{t('Number of Backups to Keep')}</ControlLabel>
                <FormControl
                  type="number"
                  value={backupType === 'number' ? settings.user.numberOfBackups : settings.user.backupDays}
                  defaultValue={30}
                  onChange={(event) => saveSetting(backupType === 'number' ? 'user.numberOfBackups':'user.backupDays', Number(event.target.value))}
                />
                <HelpBlock>{t('Backups beyond this will be erased')}</HelpBlock>
              </div>
```

You can directly add it below

murfdog64:
fyi, that was the first time I committed straight from a review.. so I might have done something noob-ish when I pulled back to my machine from git.. but that has all the changes and seems to be what we were going for

src/dashboard/components/options/OptionsHome.js
```diff
@@ -43,42 +72,35 @@ export default function OptionsHome(props) {
...
           <FormGroup controlId="backupDays">
-            <ControlLabel>{t('Number of Days of Rolling Backups to Keep')}</ControlLabel>
-            <FormControl
-              disabled
-              type="number"
-              value={settings.user.backupDays || 30}
-              onChange={(event) => saveSetting('user.backupDays', Number(event.target.value))}
-            />
-            <HelpBlock>{t('Backups older than this will be erased')}</HelpBlock>
+            {renderDropdown()}
```
jeanawhou: Looks good to me. As for my coding style, i would render it directly here so i would not need to create a function.
I prefer to separate them on a function if it has too much condition and it gets complicated like having a switch statement

Quiescent:
Ping @murfdog64.  Looks like you have conflicts to fix :(

murfdog64:
Ok @Quiescent, @jeanawhou I _think_ I have this where it needs to be this time...

Quiescent:
@murfdog64 are you sure that you've pointed at staging for remote from your local machine?
I still see it conflicting with `src/dashboard/components/options/OptionsHome.js`

Quiescent (approved):
Looks good.  LGTM!

P.S. if you want to add an example (which is a generally helpful thing to do), please put it into the examples folder.  I'm going to go ahead and move it for you to close this PR, but please remember it in the future.

PR #548, opened 2021-03-25
fixed active status to make minimap chapters readable in darkmode

murfdog64:


Quiescent:
There's a small problem here.
When we change the active chapter from `OutlineView.js` (which happens, e.g. when we edit a scene card from there).
Plottr used to update the `active` item in the minimap, received as a prop in the minimap.
Now it'll scroll to that node but not change the active item.
You could fix it by setting the state when the component updates.

murfdog64:
@Quiescent The minimap was getting the active prop, but I was struggling to get it to update which was `active` based on which was selected without the local state.
So don't set `active` in the `selectNav`, only set when the component updates?
I tried setting the state in `componentDidUpdate` in miniMap.js but it gives me the maximum update depth error..

Quiescent:
I’ve never tried replying to a GH issue via email, so let’s hope that this works! :)

I have two suggestions:
 1. Pass a callback to the child to update the state of the parent (which passes the correct value of active to the child).
 2. Use more lifecycle methods to control when it re renders.

I’m leaning towards 1. because I’m not a fan of class components and the life cycle methods.

Quiescent:
So I did some digging and here's what I think is going wrong:
 - Your callback sets it to the correct value and fires first.  Then the function finishes and the UI scrolls...
 - In `BeatView.js` we use `react-waypoint` to trigger the callback to update the active nav item.  This is where the subsequent updates come from.  It doesn't appear to fire correctly when we scroll normally (i.e. as a user using the mouse wheel).  In my experiments, it only fired when we scrolled programmatically.
 
Here's my recommendation.
The dark mode bug is the most egregious problem here and should be fixed asap.
So: I'd recommend setting `waypoint` to a dummy function on `OutlineView.js:130` and then logging a bug to fix `waypoint` in the Outline.

(Also make sure that you remove the `console.log` before pushing up again :) )

murfdog64:
Man! I saw that `waypoint` was using it too, and I figured it was related.. I shoulda kept after it!

Quiescent (approved):
Looks good!  LGTM :)

PR #558, opened 2021-03-31
Feature/kan 264

jeanawhou:
Added navbar-right padding (export button from outline view)
Screenshot from 2021-04-01 01-17-45


Quiescent (approved):
Looks good.  LGTM!

PR #559, opened 2021-03-31
kan-265 move Notes column before the Characters

jeanawhou:
move Notes column before the Characters column 
Screenshot from 2021-04-01 01-35-00


Quiescent (approved):
Looks good.  LGTM!

PR #561, opened 2021-04-01
Use different branch for templates on staging re hierarchy

Quiescent:
# Use Different Templates For Beta On/Off
## Motivation
 - We want to leave the user experience unchanged when the user turns the beta off.
 - While the beta is on, we want to leverage the new structure to create richer templates (e.g. templates with acts will have Acts on the timeline).

## Mechanics
Every time that the user changes the beta feature, we force a fetch and update of templates.

## See Also
New templates fetched from: https://github.com/Plotinator/plottr_templates/pull/3

### DIFF
--- a/src/dashboard/utils/template_fetcher.js
+++ b/src/dashboard/utils/template_fetcher.js
@@ -8,9 +8,14 @@ import {
   MANIFEST_ROOT,
 } from '../../common/utils/store_hooks'
 import { is } from 'electron-util'
+import SETTINGS from '../../common/utils/settings'
 
 const OLD_TEMPLATES_ROOT = 'templates'
 const manifestURL =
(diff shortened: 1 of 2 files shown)


src/dashboard/utils/template_fetcher.js
```diff
@@ -43,7 +48,9 @@ class TemplateFetcher {
 
   manifestReq = () => {
     return {
-      url: manifestURL,
+      // FIXME: when the beat hierarchy becomes non-beta then we can
+      // just use manifest URL.
+      url: SETTINGS.store.user.beatHierarchy ? betaManifestURL : manifestURL,
```
cameronsutter: you can also use `SETTINGS.get('user.beatHierarchy')` and it handles null/undefined

cameronsutter (commented):
Besides that little suggestion, merge it

PR #562, opened 2021-04-01
Use the old names for beat title cells when hierarchy is off

Quiescent:
# Motivation
When the Act Hierarchy is turned off, we want the user's experience to be the same as it was before.

# Change
When this PR is merged, a special case will kick in for beat titles.
If hierarchy is enabled and we're in the series view, we'll use the word "Beat" for auto-titling.
If hierarchy is enabled and we're in a book view, we'll use the word "Chapter" for auto-titling.

# Notes
When we un-beta this feature, it'll be imperative to clean up all the extra wiring.
It really is a bit cumbersome to pass the extra flags around!

### DIFF
--- a/lib/pltr/v2/helpers/beats.js
+++ b/lib/pltr/v2/helpers/beats.js
@@ -8,30 +8,58 @@ export function beatOneIsPrologue(sortedBookBeats) {
   return sortedBookBeats[0].title == i18n('Prologue')
 }
 
-export function beatName(beats, beat, sortedHierarchyLevels) {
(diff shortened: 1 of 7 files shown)


cameronsutter (approved):
@Quiescent are we ok to merge this?

Quiescent:
> @Quiescent are we ok to merge this?

Been a while since I wrote it.  It should be fine, but I'm retesting now to be sure.

Quiescent:
Looks good after testing.  LGTM!

PR #579, opened 2021-04-08
kan-268 added outfilter to redux

jeanawhou:
Added outfilter to redux
Allow the user to filter by more than one plotline
Allow user to export filtered outline

PS: I picked this ticket to learn more how plottr manages the state. I would love to know comments guys 
ticket link: https://app.ora.pm/p/4c35de9a18594ccbaee3377dcc546780?v=0&t=k&c=cc22f6e6d0ed4040a832aa0c7c78449a
Preview on slack #dev-desktop


lib/pltr/v2/helpers/cards.js
```diff
@@ -24,8 +24,13 @@ export function cardMapping(sortedBeats, sortedLines, card2Dmap, currentLine) {
 }
 
 function lineIsHidden(line, currentLine) {
-  if (!currentLine) return false
-  return line.id != currentLine
+  // if filter if previous version (1 filter id only)
+  if (Array.isArray(currentLine) && !currentLine.length) return !currentLine
+  else if (Array.isArray(currentLine) && !!currentLine.length) return !currentLine.includes(line.id)
```
Quiescent: I've picked up on two issues here:

1. The check on 28 seems redundant.  You should be able to delete it safely.
2. I think that `&& !!currentLine.length` might be redundant.
    I also feel like the extra check makes this harder to read.
    If we already know that it's an array, then whether it has a length is already checked in `.includes()`.

i.e. I think that you can rewrite 28-29 as:
```javascript
  if (Array.isArray(currentLine)) return !currentLine.includes(line.id)
```

lib/pltr/v2/helpers/cards.js
```diff
@@ -24,8 +24,13 @@ export function cardMapping(sortedBeats, sortedLines, card2Dmap, currentLine) {
...
+  // if file has no outlineFilter saved in redux
+  else if (!currentLine && !Array.isArray(currentLine)) {
```
Quiescent: If `currentLine` is falsy, then there's no need to check whether it's also not an array.

lib/pltr/v2/helpers/cards.js
```diff
@@ -24,8 +24,13 @@ export function cardMapping(sortedBeats, sortedLines, card2Dmap, currentLine) {
...
+    return false
+  } else if (!!currentLine && !Array.isArray(currentLine)) return line.id != currentLine
```
Quiescent: You can remove the `Array.isArray` from line 31 and line 33 safely.
Then this line looks much more straight forward.

src/app/components/outline/OutlineView.js
```diff
@@ -45,16 +46,18 @@ class OutlineView extends Component {
     console.warn('OutlineView waypoint needs fixing')
   }
 
-  filterItem = (id) => {
-    if (this.state.currentLine === id) {
-      this.setState({ currentLine: null })
-    } else {
-      this.setState({ currentLine: id })
-    }
+  filterItem = async (id) => {
```
Quiescent: There's no need to make this async.
We're not calling out to any APIs, and the javascript interpreter loop must finish before we do a redraw.

src/app/components/outline/OutlineView.js
```diff
@@ -45,16 +46,18 @@ class OutlineView extends Component {
...
+    const { actions, ui } = this.props
+    if (Array.isArray(ui.outlineFilter) && ui.outlineFilter.includes(id)) {
```
Quiescent: Stylistically, this is too much logic in the component.
It would be best to move this to the actions file or to the reducer (I'm leaning to the second).
The choice between the reducer and actions is made based on whether the logic is async or not.
If it's async, it goes into actions (and we'll need to add an async-aware middleware).
If it's not async (which I believe to be the case), then the reducer is the correct home for this logic.

src/app/components/outline/OutlineView.js
```diff
@@ -45,16 +46,18 @@ class OutlineView extends Component {
     console.warn('OutlineView waypoint needs fixing')
   }
 
-  filterItem = (id) => {
-    if (this.state.currentLine === id) {
```
Quiescent, line 49: It seems that your change relies on `outlineFilter` on `ui` rather than on the state.
(I don't see us calling `setState` to change `currentLine` anywhere.)
If I'm right about that, then it should probably be deleted (along with the other references to it later on).

src/app/components/outline/OutlineView.js
```diff
@@ -69,8 +72,13 @@ class OutlineView extends Component {
   }
 
   renderFilterItem(item) {
+    const { ui } = this.props
     var placeholder = <span className="filter-list__placeholder"></span>
-    if (this.state.currentLine === item.id) {
+    if (
+      (Array.isArray(ui.outlineFilter) && ui.outlineFilter.includes(item.id)) ||
+      (!Array.isArray(ui.outlineFilter) && ui.outlineFilter === item.id) ||
+      this.state.currentLine === item.id
```
Quiescent: I think this is an error?
We never set `currentLine` anymore so surely we should delete it?

src/app/components/outline/OutlineView.js
```diff
@@ -93,7 +102,11 @@ class OutlineView extends Component {
         {i18n('Outline is filtered')}
       </Alert>
     )
-    if (this.state.currentLine == null) {
+    if (
+      !ui.outlineFilter ||
+      !ui.outlineFilter.length ||
+      (!ui.outlineFilter && this.state.currentLine === null)
```
Quiescent: Same thing here re. `currentLine`.

src/app/components/outline/OutlineView.js
```diff
@@ -134,7 +147,7 @@ class OutlineView extends Component {
               beat={beat}
               cards={cardMapping[beat.id]}
               waypoint={this.fixMe}
-              activeFilter={!!this.state.currentLine}
+              activeFilter={!!ui.outlineFilter || !!this.state.currentLine}
```
Quiescent: Another reference to `currentLine`.

src/app/components/outline/OutlineView.js
```diff
@@ -143,8 +156,9 @@ class OutlineView extends Component {
   }
 
   renderBody() {
-    const { beats, lines, card2Dmap } = this.props
-    const cardMap = cardMapping(beats, lines, card2Dmap, this.state.currentLine)
+    const { beats, lines, card2Dmap, ui } = this.props
+    const { currentLine } = this.state
```
Quiescent: Same thing re. `currentLine` here.

src/common/exporter/word/exporters/outline.js
```diff
@@ -22,6 +22,7 @@ export default function exportOutline(state, namesMapping, doc, options) {
   // get current book id and select only those beats/lines/cards
   const beats = sortedBeatsByBookSelector(state)
   const lines = sortedLinesByBookSelector(state)
+  let outlineFilter = state.ui.outlineFilter
```
Quiescent: Please use `const` instead of `let` as much as you can.
`let` makes the code harder to read because the reader must search for places where `outlineFilter` could be re-assigned.

src/common/exporter/word/exporters/outline.js
```diff
@@ -44,15 +45,23 @@ export default function exportOutline(state, namesMapping, doc, options) {
...
     const title = uniqueBeatTitleSelector(state, beat.id)
     let paragraphs = [new Paragraph('')]
-    paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
+    if (!outlineFilter) {
+      paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
+    }
 
     if (options.outline.sceneCards) {
-      const cards = beatCardMapping[beat.id]
+      let cards = beatCardMapping[beat.id]
```
Quiescent: I think that this should stay `const`.
I don't see a reassignment anywhere.
Am I missing something?

src/common/exporter/word/exporters/outline.js
```diff
@@ -44,15 +45,23 @@ export default function exportOutline(state, namesMapping, doc, options) {
     const uniqueBeatTitleSelector = makeBeatTitleSelector(state)
     const title = uniqueBeatTitleSelector(state, beat.id)
     let paragraphs = [new Paragraph('')]
-    paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
+    if (!outlineFilter) {
```
Quiescent: Won't this filter out any paragraphs if there's an active filter?

src/common/exporter/word/exporters/outline.js
```diff
@@ -44,15 +45,23 @@ export default function exportOutline(state, namesMapping, doc, options) {
...
 
     if (options.outline.sceneCards) {
-      const cards = beatCardMapping[beat.id]
+      let cards = beatCardMapping[beat.id]
       const customAttrs = cardsCustomAttributesSelector(state)
       const sortedCards = sortCardsInBeat(beat.autoOutlineSort, cards, lines)
-      let cardParagraphs = sortedCards.flatMap((c) =>
-        card(c, linesById, namesMapping, customAttrs, doc, options)
-      )
+      let cardParagraphs = sortedCards.flatMap((c) => {
```
Quiescent: I think that this `let` should be a `const`.

src/common/exporter/word/exporters/outline.js
```diff
@@ -44,15 +45,23 @@ export default function exportOutline(state, namesMapping, doc, options) {
...
+        if (outlineFilter && !outlineFilter.includes(c.lineId)) {
+          return
```
Quiescent: This adds an `undefined` to the list.
Did you mean: `return []`?

src/common/exporter/word/exporters/outline.js
```diff
@@ -44,15 +45,23 @@ export default function exportOutline(state, namesMapping, doc, options) {
...
+        }
+        if (outlineFilter) {
+          paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
```
Quiescent: This side effect shouldn't happen inside of a call to `flatMap`.
Am I right in guessing that this would do the trick:
```
sortedCards.forEach((c) => {
  if (outlineFilter && !outlineFilter.includes(c.lineId))  return
  if (outlineFilter) paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
})

let cardParagraphs = sortedCards.flatMap...
```

Quiescent (changes requested):
Great approach.
I've picked up a few issues regarding correctness in the export and style in the component.

jeanawhou:
@Quiescent thank you for the very informative review!

cameronsutter (commented):
Nothing to add on top of Ed's comments

murfdog64 (commented):
I'm a little late, nothing to add 😅

jeanawhou:
@Quiescent i just updated this one. There are few conditions that i wasn't able to follow because it didn't work on my end especially this one https://github.com/cameronsutter/plottr_electron/pull/579#discussion_r609737774 i need help with this one. I pushed a working code though

Quiescent:
> @Quiescent i just updated this one. There are few conditions that i wasn't able to follow because it didn't work on my end especially this one [#579 (comment)](https://github.com/cameronsutter/plottr_electron/pull/579#discussion_r609737774) i need help with this one. I pushed a working code though

I'll take a look at the current version. :)

lib/pltr/v2/helpers/cards.js
```diff
@@ -24,8 +24,13 @@ export function cardMapping(sortedBeats, sortedLines, card2Dmap, currentLine) {
 }
 
 function lineIsHidden(line, currentLine) {
-  if (!currentLine) return false
-  return line.id != currentLine
+  // if no outlinefilter return all
+  if (Array.isArray(currentLine) && !currentLine.length) return !currentLine
```
Quiescent: I might've missed something here.  Why do you return `!currentLine` instead of `false`?

lib/pltr/v2/helpers/cards.js
```diff
@@ -24,8 +24,13 @@ export function cardMapping(sortedBeats, sortedLines, card2Dmap, currentLine) {
...
+  if (Array.isArray(currentLine) && currentLine.length) return !currentLine.includes(line.id)
+  // if file has no outlineFilter saved in redux
+  else if (!currentLine) {
+    return false
+  } else if (!!currentLine) return line.id != currentLine
```
Quiescent: This should be `else`, and please use `!==` instead of `!=` to avoid unintentional type casts.

lib/pltr/v2/reducers/ui.js
```diff
@@ -106,6 +107,16 @@ export default function ui(state = defaultUI, action) {
     case SET_TIMELINE_FILTER:
       return Object.assign({}, state, { timelineFilter: action.filter })
 
+    case SET_OUTLINE_FILTER:
+      if (action.filter === null || !action.filter) filter = null
```
Quiescent: This can be `!action.filter` it covers `null` because `null` is falsey.

lib/pltr/v2/reducers/ui.js
```diff
@@ -106,6 +107,16 @@ export default function ui(state = defaultUI, action) {
...
+      else if (Array.isArray(state.outlineFilter) && state.outlineFilter.includes(action.filter))
+        filter = state.outlineFilter.filter((item) => item !== action.filter)
+      else if (!Array.isArray(state.outlineFilter) || !state.outlineFilter) filter = [action.filter]
+      else if (!!state.outlineFilter && Array.isArray(state.outlineFilter))
```
Quiescent: We can simplify this to `Array.isArray(state.outlineFilter)` i.e. delete the `!!state.outlineFilter` because `Array.isArray` already checks for null and undefined.

src/common/exporter/word/exporters/outline.js
```diff
@@ -44,15 +45,24 @@ export default function exportOutline(state, namesMapping, doc, options) {
...
+      let flag = 0
+      const cardParagraphs = sortedCards.flatMap((c) => {
+        if (outlineFilter) {
+          if (!outlineFilter.includes(c.lineId)) return
+          flag++
+          if (flag === 1)
+            paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
+        }
+        return card(c, linesById, namesMapping, customAttrs, doc, options)
+      })
```
Quiescent: It seems to me that we want to add a level two-paragraph when we have at least one card that wasn't filtered out(?)
This code as it stands is trying to do too many things at the same time.

This has resulted in a bug: if you return from the body of a call to `flatMap` with undefined (line 59) then you'll get an undefined in an array (e.g. `[x, y, ...undefined, ..., z]`)

If we instead separate things then the problem becomes simpler.
Let's start by getting a collection of cards that aren't filtered out,
```javascript
  const filteredCards = !outlineFilter ? sortedCards : sortedCards.filter((card) => {
    outlineFilter.includes(card.lineId)
  }
```
then let's add the level 2 paragraph if there are cards,

```javascript
  if (filteredCards.length) {
    paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
  }
```

finally, let's add the paragraphs for each card.
```javascript
  const cardParagraphs = filteredCards.flatMap((c) =>
    card(c, linesById, namesMapping, customAttrs, doc, options)
  )
```

src/common/exporter/word/exporters/outline.js
Quiescent: Forget what I said before.  I see how you handled this later on.

There is a problem here.  If we have an outline filter, but `options.outline.sceneCards` is `false`, we won't end up with a level 2 paragraph (whereas before your change we used to always have a level 2 paragraph if we hit this line of code).

I think that the check should be `if (!outlineFilter || !options.outline.sceneCards)`.
WDYT?

lib/pltr/v2/helpers/cards.js
```diff
@@ -24,8 +24,13 @@ export function cardMapping(sortedBeats, sortedLines, card2Dmap, currentLine) {
 }
 
 function lineIsHidden(line, currentLine) {
-  if (!currentLine) return false
-  return line.id != currentLine
+  // if no outlinefilter return all
+  if (Array.isArray(currentLine) && !currentLine.length) return !currentLine
+  if (Array.isArray(currentLine) && currentLine.length) return !currentLine.includes(line.id)
```
Quiescent: The length check is redundant here.

Quiescent (commented):
I've gone over the PR again as it stands.
There are still a few places where the logic could be simpler and clearer.
There's also a longer (and hopefully correct this time!) response to the exporter problem.

jeanawhou:
@Quiescent i have removed the unnecessary checks, i dont know why they didn't work last time i remove them and also i hope i got the filter right. i think lodash's includes is what i missed last time. Thanks again!

src/common/exporter/word/exporters/outline.js
```diff
@@ -44,15 +45,26 @@ export default function exportOutline(state, namesMapping, doc, options) {
...
-    paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
 
     if (options.outline.sceneCards) {
       const cards = beatCardMapping[beat.id]
       const customAttrs = cardsCustomAttributesSelector(state)
       const sortedCards = sortCardsInBeat(beat.autoOutlineSort, cards, lines)
-      let cardParagraphs = sortedCards.flatMap((c) =>
-        card(c, linesById, namesMapping, customAttrs, doc, options)
-      )
+      let flag = 0
```
Quiescent: This flag seems to be redundant.  It's reset to zero whenever we enter this block anyway.
Please could you remove it?

lib/pltr/v2/helpers/cards.js
```diff
@@ -24,8 +24,13 @@ export function cardMapping(sortedBeats, sortedLines, card2Dmap, currentLine) {
 }
 
 function lineIsHidden(line, currentLine) {
-  if (!currentLine) return false
-  return line.id != currentLine
+  // if no outlinefilter return all
+  if (Array.isArray(currentLine) && !currentLine.length) return false
```
Quiescent: This line is redundant.  If it's an array then we don't need to check whether it has a length to run `.includes` on it.

lib/pltr/v2/helpers/cards.js
Quiescent: I think that this line should instead read:
```
else return line.id !== currentLine
```
because we know at this point that `currentLine` must be truthy.

lib/pltr/v2/reducers/ui.js
```diff
@@ -106,6 +107,15 @@ export default function ui(state = defaultUI, action) {
     case SET_TIMELINE_FILTER:
       return Object.assign({}, state, { timelineFilter: action.filter })
 
+    case SET_OUTLINE_FILTER:
+      if (!action.filter) filter = null
+      else if (Array.isArray(state.outlineFilter) && state.outlineFilter.includes(action.filter))
+        filter = state.outlineFilter.filter((item) => item !== action.filter)
+      else if (!Array.isArray(state.outlineFilter) || !state.outlineFilter) filter = [action.filter]
```
Quiescent: `!state.outlineFilter` is redundant here.
We can instead just write:
```
else if (!Array.isArray(state.outlineFilter))
```
because it'll already check whether it's `null`, `undefined` or `false`.

Quiescent (commented):
@jeanawhou I've found few places where I still see problems.  I hope that you find this helpful despite how long it's taking to work through these problems.
If you'd like to chat live about the changes, perhaps we can schedule some time to talk.

Quiescent:
@jeanawhou I'm going to resolve all of the issues from previous reviews to reduce noise.
You can look at the latest review for the changes that I think need to happen before merging this.

src/common/exporter/word/exporters/outline.js
```diff
@@ -44,15 +45,23 @@ export default function exportOutline(state, namesMapping, doc, options) {
     const uniqueBeatTitleSelector = makeBeatTitleSelector(state)
     const title = uniqueBeatTitleSelector(state, beat.id)
     let paragraphs = [new Paragraph('')]
-    paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
```
Quiescent, line 47: I think that there's a slight problem here.
In the case that we're not exporting scene cards, we probably still want this level 2 heading.
So can we add a check here like so:
```
if (!options.outline.sceneCards) {
  paragraphs.push(new Paragraph({ text: title, heading: HeadingLevel.HEADING_2 }))
}
```

Quiescent: @cameronsutter, could you please clarify something about this change.

I've suggested that we now need to check whether we're not exporting scene cards and if that's the case, add a level two heading.
The check would maintain what the code did before.

Bug @jeanawhou has tested the change and isn't sure what a level 2 heading looks like.

I don't want to hold up the PR if this isn't an important detail.
I'm also struggling to justify why it should remain as it was because I can't provide a good test case to see the change.

A possible test might be to check out staging and export with scene cards turned off and then switch to this branch and export again (with scene cards off) to see the difference.
WDYT?

Quiescent: I discussed and made the change with @jeanawhou.

Quiescent (approved):
Looks good.
LGTM!

PR #618, opened 2021-04-16
corrected darkmode text color for input/selected in Act Structure

murfdog64:


murfdog64:
I see now I still have the giant set of changes to package-lock... how do I avoid this?

PR #620, opened 2021-04-17
added min-height to dashboard file icon text

jeanawhou:
ticket link: https://app.ora.pm/p/613886e300e4495da43eb09e27f99a61?v=0&t=k&c=c4826140c9f244b8b2e6daff41c24a4b
I have added a min-height for the text below the dashboard icons and it still show inconsistent icon button height when on 551px size, which i think the user wouldn't want to resize the app on 550px small since it is a desktop app. What do you guys think?
dashboard-icon-text

### DIFF
--- a/src/app/components/dialogs/ActsConfigModal.js
+++ b/src/app/components/dialogs/ActsConfigModal.js
@@ -79,7 +79,9 @@ const ActsConfigModal = ({
   const RemoveLevelConfirmation = () =>
     stagedHierarchyLevels ? (
       <DeleteConfirmModal
-        customText={t('Are you sure?  (Removing levels may delete scene cards from all books across the project!)')}
+        customText={t(
+          'Are you sure?  (Removing levels may delete scene cards from all books across the project!)'
+        )}
         onDelete={() => {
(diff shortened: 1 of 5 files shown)


murfdog64 (commented):
Agreed, that fixes a lot of the janky movement of the resize up that point. Only thing they might be referring to is the auto re-sizing and wrap of flex display as the window changes size, if people aren't used to seeing that?

Quiescent:
I'm not really sure what's changed here.  Could you do a before and after?

jeanawhou:
Before the fix
dashboard-icon-inconsistent-height2
dashboard-icon-inconsistent-height

After the fix, including the image attached above which shows uneven icon text (the Open Existing) at 551px window size
icon-height-fix2
icon-height-fix1

What do you think @Quiescent ? Should i add a media query for 551px screen?

Quiescent:
I think it looks good re. heights of tiles, but (and I don't think you changed this?) the text spacing and size in the second screenshot is off.  I think that the icons and text are relatively lightweight objects and they should be farther apart than what they are right now.

Quiescent (approved):
Looks good.  LGTM!

PR #621, opened 2021-04-17
Default darkmode colors for Act Structure

murfdog64:
Went back and reapplied my changes to be more compliant with the current state of files.

Chose colors that are just lighter versions of the default lightmode colors. Ready to change them to whatever is best! What do you think?

image


cameronsutter:
hmmmm, how about we just change the background colors

All the other colors can stay the same (with one exception) but just make the background colors how you've got it right now.

The exception is the first text color. Right now in light mode the first text color is black, and then the rest follow the order of the colors for everything else. So for dark mode, let's have the first text color be light (`$darkmode-text`) and then the next ones follow the usual order of colors.

murfdog64:
Back to the original design, just fixed the one missing argument on the first round of `nextColor`
image

Quiescent:
> hmmmm, how about we just change the background colors
> 
> All the other colors can stay the same (with one exception) but just make the background colors how you've got it right now.
> 
> The exception is the first text color. Right now in light mode the first text color is black, and then the rest follow the order of the colors for everything else. So for dark mode, let's have the first text color be light (`$darkmode-text`) and then the next ones follow the usual order of colors.

I think that @murfdog64 might need to rebase and create a new file to see the change to the first colour.
I also have to say that I quite like the lighter colour palette in dark mode.

@cameronsutter what background colour were you thinking of?

I had a stab at a background colour and found that dark ones don't work well.
A light colour also needs a lighter foreground colour.
So what I ended up with here was similar to what @murfdog64 had originally (with lighter colours) but with a light gray background too.
screenshot

Quiescent:
I have to say that the light red for the first colour looks better than the light white I chose.

Quiescent:
> 
> 
> I have to say that the light red for the first colour looks better than the light white I chose.

Chatted to Cameron about this one and he said that red looks fine too :+1:

murfdog64:
Well.. I wasn't expecting that... How do I best get rid of all those commits and just go back to the very first one (27eedfc) I made?

cameronsutter (commented):
I think you're going to have conflicts with 2 other PRs, so once those get merged, you'll have to resolve those and possibly change yours a bit, but otherwise this PR is ready to go

murfdog64:
Ok, I'll keep an eye out for the conflicts

jeanawhou (approved):
Looks good to me. By the way, there's helper function to get the default text color
https://github.com/cameronsutter/plottr_electron/pull/632/files#diff-3dd5961a4ead7b30076765e90ebc9499896c4a907dae6fb6038319a9ca1d6baaR13
this can be useful in the future if this will be added but no reason to hold the PR for that.

src/app/components/timeline/BeatTitleCell.js
```diff
@@ -0,0 +1,443 @@
+import React, { PureComponent } from 'react'
```
Quiescent: So this component moved to `plottr_components`.
If you changed it in this PR, you'd need to instead make those changes to the one in `plottr_components`.

src/app/components/timeline/TimelineWrapper.js
```diff
@@ -0,0 +1,452 @@
+import React, { Component } from 'react'
```
Quiescent: So this component moved to `plottr_components`.
If you changed it in this PR, you'd need to instead make those changes to the one in `plottr_components`.

src/app/components/dialogs/ActsConfigModal.js
```diff
@@ -0,0 +1,197 @@
+import React, { useRef, useEffect, useState } from 'react'
```
Quiescent: I think you changed this component, right?
If so, you need to make those changes to the one in `plottr_components` instead.
Lemme know if you need a hand with this.

lib/pltr/v2/helpers/hierarchy.js
```diff
@@ -61,8 +62,14 @@ export const newHierarchyLevel = (allHierarchyLevels) => {
     ...hierarchyLevel,
     name: nextLevelName(allHierarchyLevels.length),
     level: 0,
-    textColor: nextColor(allHierarchyLevels.length),
-    borderColor: nextColor(allHierarchyLevels.length),
     borderStyle: nextBorderStyle(allHierarchyLevels.length),
+    dark: {
```
Quiescent, line 68: There's a slight problem here: this will cause files generated by mobile to break when opened in the desktop app.
We'll need to support both schemas for a while.
Here's a possible solution:
 - keep these keys here, and
 - copy the last set, setting, into these fields as well.

Quiescent (changes requested):
Two things:
 1. Some components which moved to `plottr_components` have re-materialised in `app`.
 2. We need to support the old schema for hierarchy level colours.
 The second one is a pain, but we really have to be careful about breaking changes in the schema of Plottr files -- we caused a lot of pain for mobile when we changed beats, and we don't want to go down that path again.

Quiescent:
When the latest review changes are addressed, I'd like to merge this and then consider what aspects of @jeanawhou 's PR are appropriate for merging.

murfdog64:
So I went through the reflog to make sure I found the changes I made, and added them to the `plottr_component` version (with the exception of `TimelineWrapper` somehow had already done that one). Funny part is now it doesn't work.. looks like I got everything this same.. but get this error

image

Went ahead and pushed so you could see again the changes, and in their new proper homes.. maybe there's something I don't know about the connection process? (thinking out loud..)

And on the revert for complying with the old schema, after some attempts, it's probably best if I take you up on your offer for assistance there too

lib/plottr_components/src/components/dialogs/HierarchyLevel.js
```diff
@@ -66,7 +67,7 @@ const HierarchyLevelConnector = (connector) => {
           setValue={setValue('textColor')}
```
Quiescent: I'm not sure that we updated this change handler to support the new schema?
You'll probably need to write a new version of `setValue` that sets the prop in the appropriate theme and sets it on the root object for backwards compatibility.  e.g.

```javascript
    const setThemedValue = (name) => (newValue) => {
      editHierarchyLevel({
        level: level,
        [name]: newValue,
        [theme]: {
          textColor,
          borderColor,
          [name]: newValue,
        }
      })
    }
```

Note that I included the old values for the `[theme]` key and then override them so that the state is merged correctly in the reducer.

Quiescent:
Did you push the changes?
When I take a look at the diff, I still see the old components re-added.
The code which supports the old schema is outlined in: https://github.com/cameronsutter/plottr_electron/pull/621/files#r634981645,  but I've written this blind so I'll leave it to you to test it and make sure that the last set value (be it for light or dark) is set to the root of the level object as well.

Quiescent (approved):
The branch now creates backwards-compatible files, but we need a follow-up PR to migrate files created from older versions/mobile.

PR #632, opened 2021-04-22
Bugfix/act structure textcolor

jeanawhou:
Gavin gray's bugfix requests
- In the Act Structure config modal, in light mode, when you remove the text color from any level, it makes the color “none”,
- Scene headings should appear as #EEE by default when you start a new project in dark mode
- In the Act Structure popup, if you click the no-color button for text color in dark mode, it should revert to #EEE

### DIFF
--- a/lib/plottr_components/src/components/dialogs/EditOrDisplay.js
+++ b/lib/plottr_components/src/components/dialogs/EditOrDisplay.js
@@ -3,9 +3,13 @@ import PropTypes from 'prop-types'
 import { DropdownButton, MenuItem, Button, Glyphicon } from 'react-bootstrap'
 import { t } from 'plottr_locales'
 import { IoIosReturnRight } from 'react-icons/io'
+import { helpers } from 'pltr/v2'
 
 import { ColorPicker, ColorPickerColor } from 'connected-components'
-import { getTextColor } from '../../../../pltr/v2/helpers/colors'
+
+const {
+  colors: { getTextColor },
+} = helpers
 
 const EditOrDisplay = ({
   id,
@@ -79,7 +83,6 @@ const EditOrDisplay = ({
   }
 
   const DisplaySwitch = () => {
-    console.log('display switch', isDarkMode)
     switch (type) {
       case 'toggle':
         return (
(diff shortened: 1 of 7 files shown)


lib/plottr_components/src/components/dialogs/EditOrDisplay.js
```diff
@@ -117,7 +118,7 @@ const EditOrDisplay = ({
               block
               title={t('No color')}
               bsStyle="warning"
-              onClick={() => setValue('none')}
+              onClick={() => setValue(`${isDarkMode ? 'none' : '#0B1117'}`)}
```
Quiescent: Can we leave this as `none` here and instead change how we render `none` later.
Otherwise, it makes things a bit more complicated than they need to be.

lib/pltr/v2/helpers/hierarchy.js
```diff
@@ -27,10 +27,11 @@ const noneIsTransparent = (borderStyle, borderColor) => {
 export const hierarchyToStyles = (
   { level, textColor, textSize, borderColor, borderStyle, backgroundColor },
   timelineSize,
-  hovering
+  hovering,
+  isDarkMode
 ) => ({
   ...{
-    color: nullIfNone(textColor),
+    color: nullIfNone(textColor === '#0B1117' && isDarkMode ? '#eee' : textColor),
```
Quiescent: Here's where I think Plottr should do all the checking.
If the colour is `none` and we're in light mode, let's use the dark colour, and we can use the light colour in dark mode.

Finally, can we put those two colours into the colours helper file in `pltr` (`lib/pltr/v2/helpers/colors.js`) and then export a function from there called `textColor` whose job it is to produce the correct text colour given the single argument `isDarkMode`.

Quiescent (changes requested):
There's just one thing I'd like us to take care of regarding whose responsibility it is to figure out text colours.
If we have small checks for the text colour like this littered throughout the repo, it'll be very difficult to change how we determine the correct text colour.

lib/pltr/v2/helpers/hierarchy.js
cameronsutter: You should also check out Colin's PR: https://github.com/cameronsutter/plottr_electron/pull/621

I think we should use his strategy. That might solve this issue as well. Or we might be able to solve this issue using his PR

Quiescent:
See: https://github.com/cameronsutter/plottr_electron/pull/670

PR #638, opened 2021-04-28
Bugfix/product activation darkmode

jeanawhou:
Ticket link: https://app.ora.pm/p/327217?c=5391647
I was not able to replicate the issue where the tabs are unclickable. So this fix is just for the verification alert to be readable. The Alert text for both error and success were unreadable
alert-danger


Quiescent (approved):
Tested it locally.
It looks good in light and dark mode.
LGTM! :)

PR #640, opened 2021-04-29
fix double entry and saving empty tags

jeanawhou:
1. fix double entry tags
2. cancel saving empty tags names
Video fix evidence will be useless since the bugs exists on keyboard events


Quiescent (approved):
Tested it locally and it looks good.
LGTM!

PR #642, opened 2021-05-04
Backup options

murfdog64:
Re-enabled backup functionality, up to choosing backup location and the display of days/number of backup (will start on full functionality for that nextimage


src/dashboard/components/options/OptionsHome.js
```diff
@@ -139,34 +141,37 @@ export default function OptionsHome(props) {
...
             {displayBackupOption()}
           </FormGroup>
         </div>
+        <hr />
+        <h1 className="secondary-text">{t('Coming Soon!')}</h1>
+        <div className="dashboard__options__item disabled">
+          <h4>{t('Auto-Save')}</h4>
+          <Switch
+            disabled
+            isOn={!!settings.user.autoSave || true}
```
jeanawhou: I think `settings.user.autoSave` will be enough

murfdog64: I didn't actually write any of that section, it moved down when I re-enabled the other parts. That's still waiting for activation another day looks like :)

src/dashboard/components/options/OptionsHome.js
```diff
@@ -35,24 +37,24 @@ export default function OptionsHome(props) {
...
+  const onChangeBackupLocation = () => {
+    const title = t('Choose your backup location')
+    const properties = ['openDirectory']
+    const files = dialog.showOpenDialogSync(win, { title, properties })
+    if (files && files.length) {
+      let folderPath = files[0]
+      setBackupLocation(folderPath)
+      saveSetting('user.backupLocation', folderPath)
+    }
+    setLoading(false)
```
Quiescent: Why do we unset loading and pre-loading here?

Quiescent:
Are these beta features?
(I'm not sure that they are because that would imply that the user could opt-out or that they're unstable.
It seems that these features are on and mandatory and that the user has no choice in using them.)

src/dashboard/components/options/OptionsHome.js
```diff
@@ -35,24 +37,24 @@ export default function OptionsHome(props) {
...
-  //   setLoading(false)
-  //   setPreloading(false)
-  // }
+  const onChangeBackupLocation = () => {
+    const title = t('Choose your backup location')
+    const properties = ['openDirectory']
+    const files = dialog.showOpenDialogSync(win, { title, properties })
+    if (files && files.length) {
+      let folderPath = files[0]
+      setBackupLocation(folderPath)
```
Quiescent: I'm not sure why we have both state and a setting for the backup location.
Looking at the way that the state for backup location works, we're setting the state inside of the render function(!)
I'm not sure that it's necessary to have state at all here, but if we need it then we should use the `useEffect` hook to trigger changes to it either on the first component mount or when some prop changes.

src/dashboard/components/options/OptionsHome.js
```diff
@@ -139,34 +141,37 @@ export default function OptionsHome(props) {
...
+              {t('Choose...')}
+            </Button>
           </FormGroup>
         </div>
-        <div className="dashboard__options__item disabled">
+        <div className="dashboard__options__item">
           <h4>{t('Days of Backup')}</h4>
           <FormGroup controlId="backupDays">
-            <select onChange={(event) => setBackupType(event.target.value)} disabled>
+            <select onChange={(event) => setBackupType(event.target.value)}>
```
Quiescent: This state is internal to the component.  i.e. if we refresh the page, the choice reverts to its default(!)
Again, I'm not sure that the state's necessary here.
We should use and alter the value in the settings store directly.

src/dashboard/components/options/OptionsHome.js
Quiescent: I spoke to @dstonem about `loading` and `preloading`, and he said that it was part of another attempt at this feature.
I'd recommend removing lines 21-38 (from where `loading` is defined up until the `useEffect`, which uses them) and then removing the `blur` and `focus` handlers in the `div` at the root of this component's render tree.

murfdog64:
https://user-images.githubusercontent.com/61074332/117364343-03a09c80-ae7b-11eb-8fd0-4abb8364e021.mov

forgot the movie

Quiescent:
Looks great and works well.  One problem: the option in the `<select>` isn't set when the page initially loads.
You can see the problem if you set it to "Number of backups" and then refresh the dashboard.
It'll show the correct value for the number of backups (and it probably has the right type of backup in the settings) but the select spuriously says "Days of backups" instead of "Number of backups".
The solution is to give the `select` a `defaultValue` or a `value` which is the value of the setting.

Quiescent:
I just tested this on Windows and found another three problems.
(Better now than when they land on prod! :) )

1. there's no default value for `backupDays`, so when you select backup days and don't change the value, and then come back, you'll find it empty.  It's only possible to test this on a fresh install, so if you want to replicate it, then be sure to delete your config files in Plottr's app data directory (I can't remember off the top of my head where that is in macOS).
2. if you try to backspace all of the input, it stops at 0.  (I'm guilty of a similar mistake in the Act Structure Modal).  The solution is to either: use a ref and an uncontrolled component or to support null when you save the value.  I think that zero isn't a valid value anyway because Plottr would better represent zero by disabling backup.
3. You can input a negative value(!)  I don't time travel, so I'm not sure what this means, but if you've found a way to reach into the future and grab new versions of books, then I want to collaborate with you on the nobel-prize-winning science! XD

Quiescent (commented):
This looks good, but I'd like to figure out what we're doing about the actual backup functionality before it lands on staging.

src/common/utils/backup.js
```diff
@@ -57,3 +65,112 @@ export function ensureBackupFullPath() {
...
 }
+
+export function deleteOldBackups(strategy, amount) {
+  const files = sortFileNamesByDate(backupFiles(BACKUP_BASE_PATH))
+
+  switch (strategy) {
+    case 'days': {
+      const anchorDate = nDaysAgo(amount)
+      const filesToDelete = files.filter((file) => !fileIsSoonerThan(anchorDate, file))
+      console.warn('`Removing old backups: ${filesToRemove}')
```
cameronsutter: let's make this a `log.warn` from `electron-log` so it shows up in the computer's log (and error reports)

src/common/utils/backup.js
```diff
@@ -57,3 +65,112 @@ export function ensureBackupFullPath() {
...
+      filesToDelete.forEach((file) => {
+        fs.unlinkSync(path.join(BACKUP_BASE_PATH, file))
```
cameronsutter: there's actually a function in electron that will send these files to the trash instead of deleting them

`shell.moveItemToTrash`

(in the newest version of electron it's: https://www.electronjs.org/docs/api/shell#shelltrashitempath)

src/common/utils/backup.js
```diff
@@ -57,3 +65,112 @@ export function ensureBackupFullPath() {
...
+      })
+      deleteEmptyFolders()
+      return
+    }
+    case 'number': {
+      const filesToDelete = files.slice(0, files.length - amount)
+      console.warn(`Removing old backups: ${filesToDelete}`)
```
cameronsutter: ditto

src/common/utils/backup.js
```diff
@@ -57,3 +65,112 @@ export function ensureBackupFullPath() {
...
+      filesToDelete.forEach((file) => {
+        fs.unlinkSync(path.join(BACKUP_BASE_PATH, file))
```
cameronsutter: ditto

cameronsutter (commented):
I just had a thought. We should probably make some sort of infinite number possible, and make that the default when this gets released. Otherwise people who have been counting on the infinite backups currently are going to be stung when it erases them after 30 days … but that does bring up something. What is going to happen when this gets released?

Quiescent:
> I just had a thought. We should probably make some sort of infinite number possible, and make that the default when this gets released. Otherwise people who have been counting on the infinite backups currently are going to be stung when it erases them after 30 days … but that does bring up something. What is going to happen when this gets released?

Yup.  That's a great point.  I'll add a default of "Never Delete Backups".

So the way that things are at the moment (without the new "Never Delete.." feature) is that the default behaviour and amount are used.
Backups that are older than 30 days are removed when the user opens a file.

Quiescent:
Thanks for the review, @cameronsutter.
Here's a summary of what I'm going to fix up:
 - minor details re. Electron,
 - infini-backup, and
 - making the whole thing async.

jeanawhou:
Looks good to me

jeanawhou (approved):

PR #643, opened 2021-05-06
Feature/font size chooser

jeanawhou:
Ticket link: https://app.ora.pm/p/334440?c=5508933

https://user-images.githubusercontent.com/19387007/117346862-e7cce300-aeda-11eb-85ea-3f96438e2e39.mp4

### DIFF
--- a/lib/plottr_components/src/components/rce/FontSizeChooser.js
+++ b/lib/plottr_components/src/components/rce/FontSizeChooser.js
@@ -54,7 +54,7 @@ export const FontSizeChooser = () => {
 }
 
 const getCurrentSize = (editor) => {
-  const [node] = Editor.nodes(editor, { match: (n) => n.font })
+  const [node] = Editor.nodes(editor, { match: (n) => n.fontSize })
   if (node) {
     return node[0].fontSize
   } else {
--- a/lib/plottr_components/src/components/rce/FontSizeChooser.js
+++ b/lib/plottr_components/src/components/rce/FontSizeChooser.js
@@ -1,13 +1,16 @@
-import React, { useState, useEffect, useRef } from 'react'
+import React, { useState, useEffect } from 'react'
+import PropTypes from 'react-proptypes'
 import { FormControl } from 'react-bootstrap'
 import { Editor } from 'slate'
-import { useSlate } from 'slate-react'
+import { ReactEditor, useSlate } from 'slate-react'
+import cx from 'classnames'
 
-export const FontSizeChooser = () => {
+export const FontSizeChooser = ({ darkMode }) => {
   const editor = useSlate()
   const [currentSize, setCurrentSize] = useState(getCurrentSize(editor))
 
   useEffect(() => {
+    ReactEditor.focus(editor)
     setCurrentSize(getCurrentSize(editor))
   }, [editor.selection])
 
@@ -32,7 +35,7 @@ export const FontSizeChooser = () => {
   return (
     <div className="font-size-chooser">
       <FormControl
-        className="btn-default"
+        className={cx({ darkmode: darkMode })}
         componentClass="select"
         onChange={changeSize}
         onSelect={changeSize}
@@ -42,6 +45,10 @@ export const FontSizeChooser = () => {
       </FormControl>
     </div>
   )
+}
+
+FontSizeChooser.propTypes = {
+  darkMode: PropTypes.bool,
 }
 
 const getCurrentSize = (editor) => {
--- a/lib/plottr_components/src/components/rce/FontSizeChooser.js
+++ b/lib/plottr_components/src/components/rce/FontSizeChooser.js
@@ -1,5 +1,5 @@
 import React, { useState, useEffect, useRef } from 'react'
-import { DropdownButton, MenuItem } from 'react-bootstrap'
+import { FormControl } from 'react-bootstrap'
 import { Editor } from 'slate'
 import { useSlate } from 'slate-react'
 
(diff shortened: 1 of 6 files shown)


Quiescent:
It looks like it works well!

In terms of the style:
The plus/minus buttons don't look like the rest of the controls, and the input's white background stands out too much.
I'd recommend making the plus/minus controls look like the other buttons and giving the input the same background as the buttons.
You could underline or outline it in orange when the user focuses the input, perhaps?

I took a look around on the web for other Rich Content Editors and found that there are two prevalent styles:
 - small up and down arrows like `^` & `v` to the right of the number, and
 - a drop-down with all the sizes offered in it.
We could consider one of those styles as well if we can't get this to work well.
I think that the drop-down looks good and probably lines up with the other tools more naturally.

lib/plottr_components/src/components/rce/FontSizeChooser.js
```diff
@@ -0,0 +1,67 @@
+import React, { useState, useEffect, useRef } from 'react'
+import { Glyphicon } from 'react-bootstrap'
+import { Editor } from 'slate'
+import { useSlate } from 'slate-react'
+
+export const FontSizeChooser = () => {
+  const editor = useSlate()
+  const fontSizeRef = useRef(null)
```
Quiescent: It seems like the font size is a controlled component because this component sets its `value`.
The ref isn't necessary because you already have the value that it's tracking.
We can safely remove it and keep track of the value with the state alone.

(Apologies if you already know this, but since there's a misstep here, I'll elaborate.  I hope that you find it useful. :)  Usually, you ought to choose between a controlled and an uncontrolled component.  You give a `value` to a controlled component and use a `ref` to get the value from an uncontrolled component.  Either approach has downsides and benefits.)

lib/plottr_components/src/components/rce/FontSizeChooser.js
```diff
@@ -0,0 +1,67 @@
...
+        }}
+      >
+        <Glyphicon glyph="minus" />
+      </button>
+      <input
+        ref={fontSizeRef}
+        className="font-size__input"
+        type="number"
+        value={currentSize || 20}
+        onChange={() => {
```
Quiescent: We could use the value from the `onChange` event here.

lib/plottr_components/src/components/rce/FontSizeChooser.js
```diff
@@ -0,0 +1,67 @@
...
+          setCurrentSize(parseInt(fontSizeRef.current.value))
+          addFontSizeMark(editor, parseInt(fontSizeRef.current.value))
+        }}
+      />
+      <button
+        disabled={currentSize > 80}
```
Quiescent: Why did we choose 81 as the maximum font size?

Quiescent (changes requested):
Looks good functionally, but I have a minor comment about how `useRef` is used in the `FontSizeChooser`.

Quiescent:
Testing it locally, I found that it had erratic behaviour on one of my files -- see the video.
Here're the problems I found:
 - The highlighted region is lost when you use the plus and minus controls.
 - Increasing and decreasing the font size doesn't seem to work in some parts of the text, but it does in others.
 
https://user-images.githubusercontent.com/1457336/117432089-3fdf0600-af2a-11eb-9357-00ab2a7f3744.mp4

lib/plottr_components/src/components/rce/FontSizeChooser.js
jeanawhou: I intended the max font size to be 80 xD tbh its just a random pick. I got it from ms paint because (if i remember it right), before the ms paint's max font size was 80.

jeanawhou:
@Quiescent i agree with the font size dropdown would be a better ux since most word processor has the same style

jeanawhou:
@cameronsutter @Quiescent @murfdog64 i have update the ui to a dropdown selector, what do you guys think? 

https://user-images.githubusercontent.com/19387007/117680584-e30e7480-b1e3-11eb-9b63-54ed99c14894.mp4

cameronsutter:
@jeanawhou I think the dropdown is a much better solution. Can you make it so a user can type a size in as well?

Let's also not forget to export the correct size when exporting to Word (Scrivener shouldn't matter in this case). I'm pretty sure that should be possible.

lib/plottr_components/src/components/rce/FontSizeChooser.js
Quiescent: Ok.  I'm not sure what it should be, so I suppose that starting with 80 and then seeing whether that will do is fine :)

Quiescent:
@jeanawhou please don't forget to take a look at the bugs I found before merging :)

Quiescent:
It looks like you have conflicts and that you've pulled in a bunch of commits that you don't need on this PR.
Could you please run `git fetch --all && git rebase -i origin/staging`.
That'll bring up your default editor (probably VIM), where you can delete the commits that you didn't make and then save the file.
Then you'll need to force push your branch back up with `git push --force-with-lease origin feature/font-size-chooser`.
When you've done that, this PR should only reflect your commits.

jeanawhou:
@Quiescent very helpful thanks a lot :)

Quiescent:
I just tested this and I couldn't replicate the bug that I pointed out before.

I'd like to merge this PR so that we have the font size picker on staging for others to test.
But I'd also like to ask for an immediate follow-up to address these two minor issues:
 1. The carret is styled differently from the font-picker (please make the carret the same as the font-picker).
 2. The default font size is 20 in the RCE, but when we export, it gets set back to the document's default (13 in the case of word).  We should either: change the default font size in Plottr's RCE or export to the same font size in Word.

Quiescent (approved):
This looks great!  LGTM!
(Please see the followup that I asked for on this PR and in Slack for two minor improvements.)

PR #648, opened 2021-05-11
Root rce normaliser

Quiescent:
# Problem
There are occasional problems with the file format, and when we fix them, we often can't hit all the files that are already in the wild.
One of the problems that we run into regularly is that the structure of the RCE is invalidated in some way.
For example, we can't nest paragraphs *inside* of other paragraphs, but the app used to do that.
The solution, until now, was that Plottr normalised the RCE content as it was edited, but this approach is problematic because we could end up with a file that has invalid RCE content *until* the user edits that content in an RCE(!)

In this PR, `pltr` accepts a new `dataRepairers` object when it constructs the root reducer.
Each reducer can pick relevant fixers off of the object and apply fixes to their content.
When the files loaded, it now normalises all of the RCE content before presenting the UI.


Quiescent:
On this branch, Plottr opens and exports these three files from the original ticket.
(I can't upload them here because GitHub won't let me.)

Quiescent:
@cameronsutter, I've pinged mobile about this one because they'd have to construct their store differently, but it looks like they're too busy to deal with this right now with their impending release.
I'm leaving this as a draft, and I'll occasionally bring it up to date to make sure that it doesn't diverge from staging too much.
If this is more important than that then I think we might need to help mobile with the change as soon as they've gotten their release out of the way -- WDYT?

Quiescent:
Merged in a different PR.

PR #651, opened 2021-05-12
Create and use a `beatIndexSelector` instead of global state

Quiescent:
# Motivation
Global state invalidates the `redux` state-management model and could result in erroneous rendering from React.
We also want to make sure that the app is insulated from the representation of state as much as possible.

This PR introduces a new selector which produces the index of a beat given the current state.


Quiescent:
Here's a recording of the UI using that selector in action:

https://user-images.githubusercontent.com/1457336/117968583-a26b4400-b326-11eb-9d89-8e7a147fc75b.mp4

cameronsutter (approved):

PR #655, opened 2021-05-14
[notes] filter functional

dstonem:


Quiescent:
The change to start using the new filter component is a little more robust and smaller in footprint so I went with that instead.

PR #658, opened 2021-05-18
tour reworked for proper state management and functionality

dstonem:


Quiescent:
@dstonem, could we please change the base for this to staging?

lib/pltr/v2/reducers/tours.js
```diff
@@ -1,5 +1,6 @@
-import { SET_TOUR_FEATURE, TOUR_NEXT, TOUR_RESTART } from '../constants/ActionTypes'
+import { SET_TOUR_FEATURE, TOUR_NEXT, TOUR_TRANSITION, TOUR_RESTART } from '../constants/ActionTypes'
 import { tour } from '../store/initialState'
+import { ACTS_TOUR_STEPS } from '../../../../src/app/components/intros/actsTourSteps'
```
Quiescent: We can't have a library depend on the app that it's embedded into.  That's a circular dependency.
We need to find another way to do this.

lib/pltr/v2/reducers/tours.js
```diff
@@ -8,56 +9,64 @@ export default function tours(state = INITIAL_STATE, action) {
     case SET_TOUR_FEATURE:
       const newState = Object.assign({}, INITIAL_STATE)
       newState.feature = action.payload //an object with feature: 'string', featureId: int
+      newState.steps = ACTS_TOUR_STEPS
```
Quiescent: If the idea is to pass the tour steps into redux when we initialise the tour, then we need to instead have these steps be part of the action.

lib/pltr/v2/reducers/tours.js
```diff
@@ -8,56 +9,64 @@ export default function tours(state = INITIAL_STATE, action) {
...
-      //delay the next step until component associated with the step is mounted
-      if (transitionSteps.includes(state.stepIndex)) {
-        state.transitioning = true
-        //need to anticipate back-to-back delays
-        if (transitionSteps.includes(state.stepIndex + 1)) {
-          state.b2bTransition = true
-        }
-      } else {
-        state.transitioning = false
+          newState.transitioning = false
```
Quiescent, line 30: Isn't this redundant because we set transitioning later?

lib/pltr/v2/reducers/tours.js
```diff
@@ -8,56 +9,64 @@ export default function tours(state = INITIAL_STATE, action) {
...
+
       //stop the tour at the appropriate endStep,
       //store that they have completed this tour,
       //clear the feature so the tour doesn't run again unless manually started
-      if (state.stepIndex === endStep) {
-        state.toursTaken[`${state.feature.name}`] = true
-        state.feature = {}
-        state.run = false
+      if (newState.stepIndex === endStep) {
+        newState.toursTaken[`${newState.feature.name}`] = true
```
Quiescent: This will mutate the previous value of state because `newState` is a shallow copy.

lib/pltr/v2/reducers/tours.js
```diff
@@ -8,56 +9,64 @@ export default function tours(state = INITIAL_STATE, action) {
...
+        newState.feature = {}
+        newState.run = false
       }
-      state.showTour = false
-      return { ...state, state }
+      
+      return { ...state, ...newState }
```
Quiescent: You shouldn't have to spread state here because `newState` has the same properties.

lib/pltr/v2/reducers/tours.js
```diff
@@ -8,56 +9,64 @@ export default function tours(state = INITIAL_STATE, action) {
...
     }
+    case TOUR_TRANSITION: 
+      const nextState = Object.assign({}, state)
+      nextState.run = false
+      nextState.loading = true
+      nextState.transitioning = false
+      return {...state, ...nextState}
```
Quiescent: You don't need to spread `state` here either.
We generally do this when we want a new version of state with just a few properties changed.

lib/pltr/v2/reducers/tours.js
```diff
@@ -8,56 +9,64 @@ export default function tours(state = INITIAL_STATE, action) {
...
     case TOUR_RESTART: {
-      const newState = INITIAL_STATE
-      return { ...state, newState }
+      const newState = Object.assign({}, state)
+      return { ...state, ...newState }
```
Quiescent: Ditto.

lib/pltr/v2/reducers/tours.js
```diff
@@ -8,56 +9,64 @@ export default function tours(state = INITIAL_STATE, action) {
     case SET_TOUR_FEATURE:
       const newState = Object.assign({}, INITIAL_STATE)
       newState.feature = action.payload //an object with feature: 'string', featureId: int
+      newState.steps = ACTS_TOUR_STEPS
       newState.showTour = true
       return { ...state, ...newState }
```
Quiescent: This is another instance of `state` being spread unnecessarily.

lib/plottr_components/src/components/timeline/TimelineWrapper.js
```diff
@@ -121,16 +121,15 @@ const TimelineWrapperConnector = (connector) => {
...
     }
 
     openBeatConfig = () => {
       this.setState({
         beatConfigIsOpen: true,
       })
       if (this.props.tour.run && this.props.tour.stepIndex === 0)
-        this.props.tourActions.tourNext('next')
+        this.props.tourActions.tourTransition()
+        setTimeout(() => this.props.tourActions.tourNext('next'),100)
```
Quiescent: I'm not sure that a timeout is the best option here.
Surely we can trigger tour next when `transitioning` becomes true -- this would be easier with hooks, but without changing this into a functional component we might be able to get away with `componentDidMount`?

lib/pltr/v2/reducers/tours.js
dstonem: It's actually more efficient to keep the steps out of redux, it's only important to store the {feature} there, and I've moved the steps back into the Tour component to avoid circular dependency and storing the entire steps array in redux. Thanks for pointing this out!

lib/pltr/v2/reducers/tours.js
dstonem: Ah so it's necessary here because if the previous step's transitioning was true then transitioning comes in as true and has to be set to false each time, but where it is redundant is in the TOUR_TRANSITION action reducer below. Removing that line now.

lib/pltr/v2/reducers/tours.js
dstonem: Changed toursTaken to an array so we avoid this mutation.

lib/plottr_components/src/components/timeline/TimelineWrapper.js
dstonem: Good call! Actioned.

PR #659, opened 2021-05-19
change select arrow icon to caret

jeanawhou:
Change the font size picker arrow down icon to `caret-down` to adapt the font picker dropdown style
font-size-picker-caret-icon


Quiescent (changes requested):
Carret looks good, but there's a minor style problem here:
The drop-down doesn't behave like a button, like the font-picker.
The drop-down also doesn't activate when you click on the carret -- that whole region should be clickable and the cursor should be `pointer` with a lightly darkened background on hover.

Hover over each to see the difference (I'm hovering my mouse over each in these screenshots, but the screenshot didn't capture the mouse):
image
image

jeanawhou:
@Quiescent thanks! i have updated the ux for the font size picker and attach a preview for it so you can check the changes.

https://user-images.githubusercontent.com/19387007/119073385-ce21b480-ba1f-11eb-9f0a-eec0cf972bb9.mp4

Quiescent:
Looks good!  LGTM :)

Quiescent:
Oops, meant to merge, not close!  Sorry!

PR #660, opened 2021-05-19
Added feature to add/remove tags, characters, places in outline

jeanawhou:
@cameronsutter I'm not sure what else is missing.

https://user-images.githubusercontent.com/19387007/118862959-eb1b9200-b910-11eb-840e-057a3ab18c17.mp4

### DIFF
--- a/lib/plottr_components/src/components/outline/CardView.js
+++ b/lib/plottr_components/src/components/outline/CardView.js
@@ -215,9 +215,6 @@ const CardViewConnector = (connector) => {
           {this.renderDropZone()}
           <div className={cx('outline-list__card-view', { darkmode: ui.darkMode })}>
             <div className="outline__card-top">
-              <div style={style} className="outline__card__line-title">
-                {line.title}
-              </div>
               <div
                 className={cx('outline__card__grip', {
                   editing: this.state.editing,
@@ -228,7 +225,10 @@ const CardViewConnector = (connector) => {
                 onDragEnd={this.handleDragEnd}
               >
                 <FaGripLinesVertical />
-                {this.state.editing ? null : <h6>{card.title}</h6>}
+                {this.state.editing ? null : <h3>{card.title}</h3>}
+              </div>
+              <div style={style} className="outline__card__line-title">
+                {line.title}
               </div>
             </div>
             <div className="outline-list__card-view__main-content">
--- a/lib/plottr_components/src/components/outline/CardView.js
+++ b/lib/plottr_components/src/components/outline/CardView.js
@@ -214,7 +214,10 @@ const CardViewConnector = (connector) => {
         >
           {this.renderDropZone()}
           <div className={cx('outline-list__card-view', { darkmode: ui.darkMode })}>
-            <div className="outline-list__card-view__left-side">
+            <div className="outline__card-top">
+              <div style={style} className="outline__card__line-title">
+                {line.title}
+              </div>
               <div
                 className={cx('outline__card__grip', {
                   editing: this.state.editing,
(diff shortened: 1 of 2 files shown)


Quiescent:
The width of left and right don't add up to 100%, but it looks like there's padding on those elements.
I wasn't able to simulate any real problems though.
When it's very narrow this happens:
image

But I don't think that's a real problem, because I'm not sure that we support such a narrow viewport.
@cameronsutter WDYT?

Quiescent:
@cameronsutter ping.  Any thoughts on this one?

cameronsutter:
re narrow viewport: I think the whole outline needs to be responsive (not within the scope of this ticket). It was made back before flexbox was even around. So let's not worry about small viewports now.

But to the more pertinent point, @jeanawhou it looks like you're not missing any functionality. I'm not in love with how it looks though. On the one hand, having it on the left like that makes sense because that's how it is in the Timeline tab, but on the other hand here it feels like it belongs on the right side. And the grab handle to drag/drop feels like it's in an odd/confusing position now.

Oh, i just had an idea. Let's put the title of the card (and the drag handle) on the top line like it was before, and then below that have the characters/places/tags on the left and the description on the right (like you've got it)

Quiescent:
I checked out the new layout, and I'm not sure about the plotline being above the title for the card.
The problem is that the title is definitely the biggest attention grabber (by size and contrast), but the hint for the plotline that it came from occupies too much of its space.
It feels awkward, and the grab handle is definitely in a weird spot.
It might work to bump it all the way to the top-right edge of the card or for it to be beneath the title.
Here's how it looks now:
screenshot

Quiescent:
I'm not sure whether this is something to address in this PR, but the visual hierarchy probably needs some work on this tab.
I feel like the card title doesn't have enough contrast with the characters, places and tags.
The bold typeface is confusing.
IMO your eyes should be drawn around in the following order:
 1. card title,
 2. card details,
 3. characters, tags etc.,
 4. plotline.

But at the moment, the characters, tags etc., draw your attention away from those others areas because they're bolded and almost as big as the title.

jeanawhou:
> I'm not sure whether this is something to address in this PR, but the visual hierarchy probably needs some work on this tab.
> I feel like the card title doesn't have enough contrast with the characters, places and tags.
> The bold typeface is confusing.
> IMO your eyes should be drawn around in the following order:
> 
> 1. card title,
> 2. card details,
> 3. characters, tags etc.,
> 4. plotline.
> 
> But at the moment, the characters, tags etc., draw your attention away from those others areas because they're bolded and almost as big as the title.

What do you think about this @cameronsutter @Quiescent ? i added the title which is an h6 with `{ font-size: 30px }` or we can use the same heading as the chapter which is an `h3`
image

Quiescent:
> What do you think about this @cameronsutter @Quiescent ? i added the title which is an h6 with `{ font-size: 30px }` or we can use the same heading as the chapter which is an `h3`
> image

I think it's better, but I'm still not sure about the plot going above the title.
What I had in mind was to put the title and plot into a row-flex box with space in between, so that the plot goes on the top right of the whole card and the title goes on the top left.
At the moment, the grab handle still looks out-of-place.

cameronsutter:
@Quiescent We used to have the plotline on the top left like you suggest, but last year sometime Ryan told us to change that to where it is now, so i don't think that'll fly unfortunately (even though that's where i would suggest it as well)

Maybe we can put it under the title. That will also add some space between the title and the characters/places/tags to hopefully help with the visual hierarchy there. Thoughts?

jeanawhou:
> Maybe we can put it under the title. That will also add some space between the title and the characters/places/tags to hopefully help with the visual hierarchy there. Thoughts?

@cameronsutter @Quiescent like this? 
image

Quiescent:
> @Quiescent We used to have the plotline on the top left like you suggest, but last year sometime Ryan told us to change that to where it is now, so i don't think that'll fly unfortunately (even though that's where i would suggest it as well)
> 
> Maybe we can put it under the title. That will also add some space between the title and the characters/places/tags to hopefully help with the visual hierarchy there. Thoughts?

I'm actually suggesting that the plotline goes in the top *right*.
Putting it under the card title is probably better, but this layout still feels a little awkward.

I had another idea that I sent Jeana over slack where we could put it into the borderline of the card, like this:
```
+-------------------------------------Plot Title---+
| || Card Title                                    |
|                                                  |
| ...                                              |
+--------------------------------------------------+
```

cameronsutter:
@Quiescent as I mentioned, i don't think either of those designs are going to fly, though I like the idea of putting it in the top right.

Let's merge it like you have it, @jeanawhou and then when I show it to Ryan we might have to tweak it a bit, but I'll handle that

PR #661, opened 2021-05-20
Tour fixes 5 17 release

dstonem:


lib/plottr_components/src/components/dialogs/ActsConfigModal.js
```diff
@@ -53,6 +54,10 @@ const ActsConfigModalConnector = (connector) => {
       }
     }
 
+    useEffect(() => {
+      tour.showTour && tourNext()
```
Quiescent, line 58: What if the user returns to the act config modal while the tour is still running?
Wouldn't that cause it to advance incorrectly?

Quiescent:
I tested out this version.  I ran into problems with transitions.
See this video:
https://user-images.githubusercontent.com/1457336/119347621-0b927600-bc9c-11eb-936c-e3d30e07b5ab.mp4

The problem is that there are assumptions made about where the event to transition comes from.
e.g. if you open the Act Structure modal a whole bunch, then it keeps on transitioning the tour, even though opening that modal should only transition from one step in the tour (and do nothing when the tour has passed that step).

I'd recommend that you turn the reducer for tours into a Finite State Machine (FSM).
Every interaction in the UI that triggers the tour would call the action with an ID, and then the tour would know what's trying to advance it.
The reducer would also be simpler in form.
It would be a nested switch statement that answers the questions:
 - "Who called me?" and
 - "What should I do about that, given where I am on the tour."

Quiescent:
There are a bunch of commits on this branch that I wouldn't expect to see.
Was the branch created from prod?
If it was then it's probably best to interactively rebase it onto staging and drop the commits that aren't specific to this branch with `git fetch --all && git rebase --interactive origin/staging`

lib/plottr_components/src/components/dialogs/ActsConfigModal.js
dstonem: Going to implement a quick fix in React while I look into Finite State Management - in case we want to release a patch in the next day or two with the tour included. Thanks for catching this!

PR #662, opened 2021-05-24
Make Acts Tour More Robust

dstonem:
After the rebasing from `tour-fixes-5-17-release` to remove extraneous commits didn't work, I made a new branch from the current `staging` and inserted the updated tour code. The important change here is the conditional calling of `tourNext()` in the `ActsConfigModal` `useEffect` on line 58. I have not been able to recreate any of @Quiescent's errors noted here: https://user-images.githubusercontent.com/1457336/119347621-0b927600-bc9c-11eb-936c-e3d30e07b5ab.mp4

My thoughts being we can release the tour in any upcoming patch releases as-is prior to shifting the tour functionality to a Fixed State Machine for greater simplicity in the tour reducer.

re: https://github.com/cameronsutter/plottr_electron/pull/661#issuecomment-847012249

### DIFF
--- a/lib/pltr/v2/reducers/tours.js
+++ b/lib/pltr/v2/reducers/tours.js
@@ -1,65 +1,142 @@
-import { SET_TOUR_FEATURE, TOUR_NEXT, TOUR_RESTART } from '../constants/ActionTypes'
+import {
+  SET_TOUR_FEATURE,
+  TOUR_NEXT,
+  TOUR_RESTART,
+  TOUR_PREVIOUS,
+  RESET_ADVANCE_SOURCE,
+  SET_STEP,
+} from '../constants/ActionTypes'
 import { tour } from '../store/initialState'
+import { ACTS_TOUR } from '../constants/tour'
 
 const INITIAL_STATE = tour
 
 const tours = (dataRepairers) => (state = INITIAL_STATE, action) => {
+  const { feature, source, appInteractionAdvanced, overridenStepIndex, resetIfMismatch } = action
+
   switch (action.type) {
     case SET_TOUR_FEATURE: {
-      const newState = Object.assign({}, INITIAL_STATE)
-      newState.feature = action.payload //an object with feature: 'string', featureId: int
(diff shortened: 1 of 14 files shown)


lib/pltr/v2/reducers/tours.js
```diff
@@ -9,57 +9,61 @@ export default function tours(state = INITIAL_STATE, action) {
...
+
       //stop the tour at the appropriate endStep,
       //store that they have completed this tour,
       //clear the feature so the tour doesn't run again unless manually started
-      if (state.stepIndex === endStep) {
-        state.toursTaken[`${state.feature.name}`] = true
-        state.feature = {}
-        state.run = false
+      if (newState.stepIndex === endStep) {
+        newState.toursTaken.length ? newState.toursTaken = [...newState.toursTaken,newState.feature.name] : [newState.feature.name]
```
Quiescent: It seems like we don't use the toursTaken property (I searched the repo with your branch checked out).
Is that intentional?

Quiescent:
I can't replicate the transition bug anymore -- great! :)

Here are two more problems:
 1. If the user doesn't actually increase the levels of hierarchy, the tour gets stuck.
 2. If the user closes the act structure modal and hovers of the dot, the app crashes because the component that the dot was attached to dissapears.

Here's a recording for 1: 

https://user-images.githubusercontent.com/1457336/119614380-82478480-bdfe-11eb-86ca-0aa4176687a1.mp4

Finally, I might have misunderstood the intention of the original PR.
If this is intended to be a patch release then we do actually need to base it off of the release branch.
(Staging is not in a good state to cut a release branch at the moment because it lacks a migration for the dark/light theme change).

Quiescent:
@cameronsutter @jeanawhou I've fixed up a few issues in the tour and implemented what Dylan and I discussed.
The tour should be more resilient to the user not doing what was expected now.
Can you please try to break the tour -- it should be much harder to do now.

Quiescent:
Defunct.

PR #663, opened 2021-05-27
Next migration

murfdog64:


lib/pltr/v2/migrator/migrations/2021.6.9.js
```diff
@@ -0,0 +1,22 @@
+const { cloneDeep } = require('lodash')
+
+function migrate(data) {
+  if (data.file && data.file.version === '2021.5.xx') return data
```
Quiescent: Please make this 2021.6.9.

lib/pltr/v2/migrator/migrations/__tests__/2021.6.9.js
```diff
@@ -0,0 +1,31 @@
...
+  describe('given the initial file state', () => {
+    it('should produce the initial file state', () => {
+      expect(migration(initialFile)).toEqual(initialFile)
+    })
+  })
+  describe('given a state with hierarchyLevel in it', () => {
+    const { hierarchyLevel } = require('../../../store/initialState')
+    const stateWithHierarchy = {
+      ...initialFile,
+      hierarchyLevels: { hierarchyLevel },
```
Quiescent: This schema isn't correct.  It should be: `{ 0: hierarchyLevel }`.
Curiously, the test will still work because the migration doesn't care what the keys are called.
It's important to correct this because people use tests as documentation for how things are supposed to work.

lib/pltr/v2/migrator/migrations/__tests__/2021.6.9.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+    it('should effect a change of any nature', () => {
+      expect(migration(stateWithHierarchy)).not.toEqual(stateWithHierarchy)
+    })
+    it('should produce a hierarchy with dark and light attributes', () => {
+      const migrated = migration(stateWithHierarchy)
+      expect(migrated.hierarchyLevel.dark).toEqual({ borderColor: '#c9e6ff', textColor: '#c9e6ff' })
```
Quiescent: I expect this to fail(!)
The assertion should be: `expect(migrated.hierarchyLevels.hierarchyLevel.dark).toEqual({ borderColor: '#c9e6ff', textColor: '#c9e6ff' })` (and when you change the key to be zero then ofcourse this assertion will change again.

Did you run the test?
You can run it with `npm run test`.

Quiescent (changes requested):
Looks good, but I found a problem with the test and the name of the migration in the check to not re-migrate.

Quiescent (approved):
Looks good.  LGTM!

PR #664, opened 2021-05-27
Feature/tags categories

jeanawhou:
Ticket link: https://app.ora.pm/p/334440?c=5508960
I also have added an opacity of `50%` for the category name when the wrapper is hovered. What do you think guys?

https://user-images.githubusercontent.com/19387007/119874821-9acbb200-bf58-11eb-9f07-38b7c81c23a1.mp4


Quiescent:
Looks great so far.  There are two things that I'd like to address before merging:
 1. When you stop hovering over a tag, it looks like the title remains at the lower level of opacity.  I think that it should return to its original opacity when the controls disappear.  If that's too difficult then I recommend that we don't fiddle with the opacity at all.
 2. We've removed the toolbar "Add Tag" button.  @cameronsutter I think that this the best option in this case because it's not clear where a tag should be created if it's outside of a categroy.  WDYT?

lib/plottr_components/src/components/tag/TagCategoriesModal.js
```diff
@@ -0,0 +1,86 @@
...
+
+  function TagCategoriesModal({
+    categories,
+    ui: { darkMode },
+    closeDialog,
+    addTagCategory,
+    deleteTagCategory,
+    updateTagCategory,
+    reorderTagCategory,
+    startSaveAsTemplate,
```
Quiescent: Are we planning on template functionality for tags?

lib/plottr_components/src/components/tag/TagListView.js
```diff
@@ -3,44 +3,141 @@ import PropTypes from 'react-proptypes'
...
 
   class TagListView extends Component {
     state = {
       appending: false,
+      categoriesDialogOpen: false,
+      tagDetailId: null,
+      newCategoryId: null,
+    }
+
+    static getDerivedStateFromProps(props, state) {
```
Quiescent, line 23: This and the next static function should be selectors.  (That way other apps, like mobile, can use the same wiring to get tag categories etc.)

Quiescent:
Other than my two comments above, this looks good.

jeanawhou:
> Looks great so far. There are two things that I'd like to address before merging:
> 
> 1. When you stop hovering over a tag, it looks like the title remains at the lower level of opacity.  I think that it should return to its original opacity when the controls disappear.  If that's too difficult then I recommend that we don't fiddle with the opacity at all.

Yeah i realized that it doesn't look that bad though the action buttons cover the category name that's why i reduced the opacity from 10 - 50%. I will just the remove the onhover opacity

lib/plottr_components/src/components/tag/TagCategoriesModal.js
jeanawhou: This is not included in the ticket, i will just remove this extra code

cameronsutter:
> 2\. We've removed the toolbar "Add Tag" button.  @cameronsutter I think that this the best option in this case because it's not clear where a tag should be created if it's outside of a categroy.  WDYT?

yeah that's fine

One tweak though that I'd like to make is to the styling of each category:
- Let's remove the border and background color around each category
- and reduce the font size of the category name to the same as it is in Characters (we may have to make it a bit bigger, but i at least want to see how that looks)

jeanawhou:
> * Let's remove the border and background color around each category
> * and reduce the font size of the category name to the same as it is in Characters (we may have to make it a bit bigger, but i at least want to see how that looks)

@cameronsutter this is how it'll looked. The category name is an h2 with a 17px font-size
Screen Shot 2021-05-28 at 11 33 41 PM

cameronsutter:
Looks perfect!

Quiescent:
It looks great!

I checked it out locally and tested it this morning.  Here are two small quirks that I think are worth addressing before we merge:
 1. When categories are added, there's no add button for the new category.  You have to move a tag to the new category to make it show up.  It would be better if we always rendered the add button, even when a category is empty.
 2. The layout shifts when new categories are rendered.  See attached video.
 
https://user-images.githubusercontent.com/1457336/120158556-8c012b80-c1f4-11eb-93ec-782eb70450a3.mp4

jeanawhou:
> It looks great!
> 
> I checked it out locally and tested it this morning. Here are two small quirks that I think are worth addressing before we merge:
> 
> 1. When categories are added, there's no add button for the new category.  You have to move a tag to the new category to make it show up.  It would be better if we always rendered the add button, even when a category is empty.
> 2. The layout shifts when new categories are rendered.  See attached video.
> 
>  2021-05-31-09.48.48.mp4

What do you think about this @Quiescent @cameronsutter ?

image

Quiescent:
> What do you think about this @Quiescent @cameronsutter ?
> 
> image

I'm not sure about centre-aligning the "Tags" title if the category titles are left-aligned.
But I'm equally unsure about how it might look with left alignment, so my verdict is: let's merge it and see what feedback we get once it's on staging.

Quiescent (approved):
I think that this looks great.
We might find more when it lands on staging, but I'm happy with how it looks now.

PR #665, opened 2021-05-28
[timeline] preview lines for drag-n-drop

dstonem:


lib/plottr_components/src/components/timeline/BeatTitleCell.js
```diff
@@ -36,6 +37,9 @@ const BeatTitleCellConnector = (connector) => {
         inDropZone: false,
         dropDepth: 0,
         deleting: false,
+        draggedBeat: {},
```
Quiescent: I don't think that you actually use this piece of state.

lib/plottr_components/src/components/timeline/BeatTitleCell.js
```diff
@@ -154,9 +158,15 @@ const BeatTitleCellConnector = (connector) => {
       if (event.which === 27) this.setState({ editing: false })
     }
 
+    handleSetDraggedBeat = beat => {
```
Quiescent: This is unnecessary lambda wrapping.
You can use `this.props.setDraggedBeat` wherever you were using this function instead.

Quiescent:
Please run prettier and `eslint` on your code.
I don't use VSCode, but there is an EsLint plugin that should help you to keep your code in line with our standards: https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

lib/plottr_components/src/components/timeline/TimelineTable.js
```diff
@@ -32,6 +32,7 @@ const TimelineTableConnector = (connector) => {
     state = {
       tableLength: 0,
       hovering: null,
+      draggedBeat: {}
```
Quiescent: Are you sure that this is the best way to encode that there's no beat being dragged currently?
If I wanted to check that a beat was being dragged, I'd need to check for an empty attribute on this object rather than checking whether there is an object at all.
(I'd recommend using `null` instead here).

lib/plottr_components/src/components/timeline/TopRow.js
```diff
@@ -23,6 +23,7 @@ const TopRowConnector = (connector) => {
       super(props)
       this.state = {
         hovering: null,
+        draggedBeat: {}
```
Quiescent: Ditto for the absence of a beat being dragged issue.

Quiescent:
This issue is very stale and it would be better to start from scratch when we know what we want.

PR #668, opened 2021-06-02
Create a normalize function for RCEs

Quiescent:
# Problem
There are occasional problems with the file format, and when we fix them, we often can't hit all the files that are already in the wild.
One of the problems that we run into regularly is that the structure of the RCE is invalidated in some way.
For example, we can't nest paragraphs *inside* of other paragraphs, but the app used to do that.
The solution, until now, was that Plottr normalised the RCE content as it was edited, but this approach is problematic because we could end up with a file that has invalid RCE content *until* the user edits that content in an RCE(!)

In this PR, `pltr` accepts a new `dataRepairers` object when it constructs the root reducer.
Each reducer can pick relevant fixers off of the object and apply fixes to their content.
When the files loaded, it now normalises all of the RCE content before presenting the UI.


Quiescent:
Validate passed locally.

PR #673, opened 2021-06-13
Feature/export customize plotlines

jeanawhou:
ticket link: https://app.ora.pm/p/320746?c=5254986
changes preview (youtube link) [https://youtu.be/yEqZ4cjnydI](url)

i didnt expect that quicktime recorder will bloat the video record to 380mb, i just uploaded preview to youtube cuz i still dont have a light weight screen recorder to use. 

What do you think guys ? See video link attach


Quiescent:
> i didnt expect that quicktime recorder will bloat the video record to 380mb, i just uploaded preview to youtube cuz i still dont have a light weight screen recorder to use.

You can use [HandBrake](https://handbrake.fr/) to encode it.  Using the Web 1080p preset I've found that it usually reduces the size to 1-3Mb depending on the length of the recording.

Quiescent:
Just a general note about keeping your branches up to date with `staging`: please rebase your branches instead of merging `staging` into them.
That way, you'll avoid the merge commits.
You can squash them and synchronize your branch in one go by running `git fetch --all && git rebase --interactive origin/staging`.
I'd recommend that you squash these commits into one because something's gone horribly wrong here (the editor that pops up when you do an interactive rebase lets you squash commits by changing the command at the start of the relevant lines).
I would never expect to see two commits, followed by two merges of staging into your branch, followed by those same commits again(!)

Quiescent:
Functionally, it's looking great! :)
I'm going to do a once-over on the code to make sure that we're still good to go there.

src/common/exporter/scrivener/v2/exporters/beats.js
```diff
@@ -28,15 +28,23 @@ export default function exportBeats(state, documentContents, options) {
   const linesById = keyBy(lines, 'id')
   const customAttrs = cardsCustomAttributesSelector(state)
   const outlineFilter = state.ui.outlineFilter
+  const outlineExportFilter = options.outline.filter
```
Quiescent, line 31: Isn't this the same as the `outlineFilter`?
(When would they be different?)

src/common/exporter/word/exporters/outline.js
```diff
@@ -23,6 +23,7 @@ export default function exportOutline(state, namesMapping, doc, options) {
   const beats = sortedBeatsByBookSelector(state)
   const lines = sortedLinesByBookSelector(state)
   const outlineFilter = state.ui.outlineFilter
+  const outlineExportFilter = options.outline.filter
```
Quiescent, line 26: Same question as above.

Quiescent:
Prettier seems to disagree about the way that `ui.js` changed formatting.
Please check the log for the GH action here: https://github.com/cameronsutter/plottr_electron/pull/673/checks?check_run_id=2814643981

src/common/exporter/scrivener/v2/exporters/beats.js
jeanawhou: @Quiescent these are the lines that are checked inside the `Advanced export options` modal when you click the export button then choose `Advanced..` 

They will differ once the FilterCheckOption is changed, it will push the `lineId` to the `options.outline.filter`. 

If there's an `outlinefilter` those `lineId` will be selected by default, and if the user select more plotlines, all those ids (including the `outlineFilter`) will be stored to `options.outline.filter`.

`options.outline.filter` is empty if the user will not click the `Advanced` export option and will only export the the `lineId` from the `outlineFilter` from the store, if no outline filter then the entire outline will be exported.

jeanawhou:
> Prettier seems to disagree about the way that `ui.js` changed formatting.
> Please check the log for the GH action here: https://github.com/cameronsutter/plottr_electron/pull/673/checks?check_run_id=2814643981

Im not sure what's wrong with my prettier, I tried to use the node version same what's on the `nvmrc` and used `npm` instead of the `yarn` (im using yarn because i always encounter issues when i use npm :") ) but it formats differently from yours. 

No only on `ui.js` but also on these files. (i just excluded them from the commit)
image

src/common/exporter/scrivener/v2/exporters/beats.js
Quiescent: Cool.  That sounds sensible.

One question: does the outline filter (when you're done exporting and head back to the outline) change according to the options that you selected in the exporter?

Quiescent:
> > Prettier seems to disagree about the way that `ui.js` changed formatting.
> > Please check the log for the GH action here: https://github.com/cameronsutter/plottr_electron/pull/673/checks?check_run_id=2814643981
> 
> Im not sure what's wrong with my prettier, I tried to use the node version same what's on the `nvmrc` and used `npm` instead of the `yarn` (im using yarn because i always encounter issues when i use npm :") ) but it formats differently from yours.
> 
> No only on `ui.js` but also on these files. (i just excluded them from the commit)
> image

That's very strange!
There are two things that we need to address here:
 1. What's going wrong with `npm`?  (This project uses `npm` and there are important differences between `npm` and `yarn`.  If something's going wrong then we should try to fix it sooner rather than later.)  Also, what version of `npm` are you running (it should be version 7).
 2. What command are you running to format the project?  Do you have a `.prettierrc` file in your home directory?  (That's usually at `/home/<username>` on Linux and `/Users/<username>` on MacOS).

src/common/exporter/scrivener/v2/exporters/beats.js
jeanawhou: The `outlineFilter` from store will be the same, but the `options.outline.filter` from the `Advanced export` modal will be set to null since the number of plotlines can vary but other settings are saved by default (same on prod).

jeanawhou:
> Just a general note about keeping your branches up to date with `staging`: please rebase your branches instead of merging `staging` into them.
> That way, you'll avoid the merge commits.
> You can squash them and synchronize your branch in one go by running `git fetch --all && git rebase --interactive origin/staging`.
> I'd recommend that you squash these commits into one because something's gone horribly wrong here (the editor that pops up when you do an interactive rebase lets you squash commits by changing the command at the start of the relevant lines).
> I would never expect to see two commits, followed by two merges of staging into your branch, followed by those same commits again(!)

@Quiescent I encountered problem rebasing, my latest code will be overwritten from my previous commits.

src/common/exporter/scrivener/v2/exporters/beats.js
Quiescent: I'm not sure that I understand.  Are you saying that the `outlineFilter` in the store is set to the same value as what you choose in the `Advanced Export`, but that the `options.outline.filter` is set to `null` every time that we export?
What happens to `options.outline.filter` if the user turns on the "Save these settings across projects" setting?

Quiescent:
> @Quiescent I encountered problem rebasing, my latest code will be overwritten from my previous commits.

I'm not sure that I understand.  Are you saying that your previous commits will take precedence over your recent commits?
That should never happen.  Are you available to hop on a call at some point today to go through this?
Does 4 PM SAST work for you?

src/common/exporter/scrivener/v2/exporters/beats.js
jeanawhou: > I'm not sure that I understand. Are you saying that the outlineFilter in the store is set to the same value as what you choose in the Advanced Export, but that the options.outline.filter is set to null every time that we export?

Yes, the `outlineFilter` in the store (if there is) will be the selected values of the `options.outline.filter`, if `outlineFilter` is null, then we select all the `lineIds` to be exported.

If the user select/check another `plotline` to be exported, the `outlineFilter` values will be mapped to the `options.outline.filter` plus the new `plotline` that was selected. 

line 34 will be true
```
if (outlineExportFilter) {      return cards.filter((card) => includes(outlineExportFilter, card.lineId))    } 
```

> but that the options.outline.filter is set to null every time that we export?
> What happens to options.outline.filter if the user turns on the "Save these settings across projects" setting?

Yes `options.outline.filter` will always be null every after export and they will not be saved even if the user turns on the "Save these settings across projects" setting
Because each story has different number of plotlines / `lineId`, the export will be broken if we will save the line especially when the previous pltr file has more plotlines

PR #678, opened 2021-06-17
added export word category option

jeanawhou:
Ticket link: https://app.ora.pm/p/320746?c=5255022
@cameronsutter @Quiescent im not sure why im not able to access the `categoryHeading` i added on the `default_config.js`
what i did is 
```
if (category is selected from the advanced export options and has category) {
  render category as Heading 
  render category Name
}
```

https://user-images.githubusercontent.com/19387007/122443443-2ec9f000-cfd2-11eb-9a0b-1ab9338d5c9f.mp4


src/common/exporter/word/exporters/characters.js
```diff
@@ -49,6 +54,14 @@ function characters(characters, customAttributes, images, doc, options) {
     if (options.characters.description) {
       paragraphs.push(new Paragraph(ch.description))
     }
+    if (options.characters.category && ch.categoryId) {
+      paragraphs.push(new Paragraph({ text: t('Category'), heading: HeadingLevel.HEADING_3 }))
```
Quiescent, line 58: This should be the plural: "Categories"

Quiescent:
> Ticket link: https://app.ora.pm/p/320746?c=5255022
> @cameronsutter @Quiescent im not sure why im not able to access the `categoryHeading` i added on the `default_config.js`
> what i did is
> 
> ```
> if (category is selected from the advanced export options and has category) {
>   render category as Heading 
>   render category Name
> }
> ```
> 
> vokoscreen-2021-06-18_01-04-50.mp4

@jeanawhou, I think that the problem is that the config is saved to disk.
I tested this theory and found that it was correct:
Default config has the value,
image

but it's overwritten by the config on-disk:
image

So when we change the default config it doesn't affect what's stored on the disk.
May I recommend that you merge the disk config with the defaults and then set the value of the store on `store_hooks.js:117`.
You may want to check out:
 - https://github.com/sindresorhus/electron-store#defaults, and
 - https://github.com/sindresorhus/electron-store#setobject.
Something like this might work:
```javascript
import { merge } from 'lodash'
...

exportConfigStore.set(merge(exportConfigStore.defaults, exportConfigStore.options))
return useJsonStore(exportConfigStore)
```

Quiescent:
Found a bug while testing this morning.
If a character isn't categorised Plottr blows up when you export because the category is `undefined` for uncategorised characters:
image

cameronsutter (approved):

PR #680, opened 2021-06-18
Remove unneeded packages

Quiescent:
# Motivation: To Clean Up Unreferenced Dependencies
This isn't exhaustive.
I spent a minute grepping for references and found that we probably don't use these.
Here's a summary from my search:

```
 - [X] electron-prompt
 - [X] acorn (parser, don't see where it's used)
 - [X] ajv (validation generator for json schemas)
 - [X] devtron (a dev tool extension which doesn't seem to load)
 - [X] electron-devtools-installer (installs devtools, but it doesn't work with our version of electron!)
 - [X] electron-reload spuriously included along with the next package
 - [X] electron-reloader (used to work, but it's been turned off for ages and we've been just fine without it)
 - [X] react-overlays (seems to be a transitive dependency that we installed directly)
 - [ ] request (deprecated.  We should use axios intsead)
```


cameronsutter:
@Quiescent I'm fine removing most of these. I'm unsure about 2 of them though.

Acorn and ajv were added because either github's dependabot added them explicitly to bump the version number for a security patch or they were a peer dependency of one of our packages that npm said was missing. I can't remember which was the case, but let's test without these two and see if they truly aren't needed.

Quiescent:
> @Quiescent I'm fine removing most of these. I'm unsure about 2 of them though.
> 
> Acorn and ajv were added because either github's dependabot added them explicitly to bump the version number for a security patch or they were a peer dependency of one of our packages that npm said was missing. I can't remember which was the case, but let's test without these two and see if they truly aren't needed.

Noted.  One question: when you say, "let's test without these two" do you mean merge it into staging and test from there?

cameronsutter:
@Quiescent yup. Let's merge it

PR #681, opened 2021-06-18
update dashboard activation darkmode text

jeanawhou:
Ticket link: https://app.ora.pm/p/320746?c=5230154
@cameronsutter @Quiescent  I have tested product activation and i don't see illegible text other than the `Days Left` and `License` or the activation link.

Before
activation-dark-before

Update
default-link-color


Quiescent:
@jeanawhou The update looks better.
I might've gone for an even lighter colour, but this is definitely legible now.

I have a problem with where the changes to the style sheets were made: we can't change `bootstrap`.
That's an embedded library that we don't upstream to.
We need to override its styles in other files.

cameronsutter:
The style changes look fine.

I agree that we shouldn't change the bootstrap files themselves. So you'll have to find the right place to override those styles in our stylesheets.

Quiescent (approved):
This is much better!  Good job, LGTM! :)

PR #682, opened 2021-06-18
fix mini-color-picker position

jeanawhou:
ticket link: https://app.ora.pm/p/320746?c=5523838

preview

https://user-images.githubusercontent.com/19387007/122593292-073d5b00-d098-11eb-891e-06faea31e5a6.mp4

### DIFF
--- a/lib/plottr_components/src/components/MiniColorPicker.js
+++ b/lib/plottr_components/src/components/MiniColorPicker.js
@@ -49,7 +49,7 @@ export default function MiniColorPicker(props) {
   return (
     <div
       className={cx('mini-color-picker', { darkmode: props.darkMode })}
-      style={useCoords()}
+      style={props.position || useCoords()}
       ref={pickerRef}
     >
       <p>{i18n('Default Colors')}</p>
@@ -133,4 +133,5 @@ MiniColorPicker.propTypes = {
   close: PropTypes.func,
   chooseColor: PropTypes.func,
   darkMode: PropTypes.bool.isRequired,
+  position: PropTypes.object,
 }
--- a/lib/plottr_components/src/components/timeline/CardDialog.js
+++ b/lib/plottr_components/src/components/timeline/CardDialog.js
@@ -508,7 +508,7 @@ const CardDialogConnector = (connector) => {
             <ColorPickerColor
               color={card.color || card.color === null ? 'none' : '#F1F5F8'} // $gray-9
               choose={this.openColorPicker}
-              style={{ margin: '2px', marginRight: '12px', height: 50, width: 55 }}
+              style={{ margin: '2px', marginRight: '12px' }}
               buttonStyle={{
                 border: `1px solid ${borderColor}`,
                 backgroundColor: card.color && borderColor,
@@ -539,6 +539,7 @@ const CardDialogConnector = (connector) => {
                 chooseColor={this.chooseCardColor}
                 el={this.colorButtonRef}
                 close={() => this.setState({ showColorPicker: false })}
+                position={{ left: 118 }}
               />
             </Overlay>
           </div>
--- a/lib/plottr_components/src/components/timeline/CardDialog.js
+++ b/lib/plottr_components/src/components/timeline/CardDialog.js
@@ -508,7 +508,7 @@ const CardDialogConnector = (connector) => {
             <ColorPickerColor
               color={card.color || card.color === null ? 'none' : '#F1F5F8'} // $gray-9
               choose={this.openColorPicker}
-              style={{ margin: '2px', marginRight: '12px' }}
+              style={{ margin: '2px', marginRight: '12px', height: 50, width: 55 }}
               buttonStyle={{
                 border: `1px solid ${borderColor}`,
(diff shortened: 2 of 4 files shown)


Quiescent (changes requested):
Looks good, except for one problem: the picker and no-colour controls are bumped when you open the picker.
Could you please take a look at preventing that from happening?

cameronsutter:
I think we'll need a parameter sent to MiniColorPicker that switches whether it's rendering in the RCE toolbar or not. The position it calculates is normally based on the RCE toolbar, which will be wrong in the case of the card dialog

jeanawhou:
@cameronsutter @Quiescent i think the problem here is the `ColorPickerColor` button as it auto expands when the `MiniColorPicker` is visible

What do you think of this `height` and `width` override? I set a fixed height and width for the `ColorPickerColor` button inside the card dialog
image

Quiescent:
> What do you think of this `height` and `width` override? I set a fixed height and width for the `ColorPickerColor` button inside the card dialog

I think that we've gone for pixel-defined widths a bit too much.
These two components are siblings.
Perhaps we could try putting the `MiniColorPicker` overlay (which is at the end of the DOM tree as a sibling of the `ColorPickerColor`) vertically beneath its current siblings using a column-based flex layout.

i.e. something like:
```javascript
    <div className="color-picker__container"> {/* <-- add this as a column arranged flex container */}
      <div className="color-picker__box">
            <label>
              ...
            </label>
            <ColorPickerColor
               ...
             >
             ...
            </div>
     </div>
    {/* Now the overlay isn't in the same div, and should be vertically beneath the controls without impacting their layout. */}
      <Overlay
         ...
      >
      <MiniColorPicker
         ...
      />
     </Overlay>
   </div>
```

jeanawhou:
@cameronsutter  i have tried implementing adding a position if the `<MiniColorPicker />` is in the `CardDialog.js` 
the position varies depending on the contents of the characters, places and tags

More right side contents
image

Lesser right side contents
image


@Quiescent i have also tried adding a `color-picker__container` but the `MiniColorPicker` will be appended to its container thus making it none effect, and still pushed the `no-color-button` to the right

image

jeanawhou:
@Quiescent i tried to add `color-picker__container` to wrap the `color-picker__box` and the `Overlay` but it appended the MiniColorPicker to its container, thus making it to no effect and still pushed the `no-color` button to the right. 

image

What i did is i followed what @cameronsutter advise to add a prop which is the `position`. If there's no position then we use the `useCoords` hook from the MiniColorPicker

cameronsutter:
@jeanawhou hmmmm, did you also move the `<Overlay>` out of `color-picker__box` when you did that?

The `position` prop looks good

jeanawhou:
@cameronsutter  I made the `color-picker__container` parent of both `color-picker__box` and the `<Overlay />` 

The `color-picker__box` wasn't move I only wrap it to be siblings with overlay

 This is the entire code block that i tried
```
<div className="color-picker__container">
  <div className="color-picker__box">
    <label className="card-dialog__details-label" style={{ minWidth: '55px' }}>
      {i18n('Color')}:
    </label>
    <ColorPickerColor
      color={card.color || card.color === null ? 'none' : '#F1F5F8'} // $gray-9
      choose={this.openColorPicker}
      style={{ margin: '2px', marginRight: '12px' }}
      buttonStyle={{
        border: `1px solid ${borderColor}`,
        backgroundColor: card.color && borderColor,
      }}
      ref={this.colorButtonRef}
    />
    <div className="buttons" style={{ alignSelf: 'flex-start' }}>
      <Button bsSize="xs" block title={i18n('Choose color')} onClick={this.openColorPicker}>
        <Glyphicon glyph="tint" />
      </Button>
      <Button
        bsSize="xs"
        block
        title={i18n('No color')}
        bsStyle="warning"
        onClick={() => this.chooseCardColor(null)}
      >
        <Glyphicon glyph="ban-circle" />
      </Button>
    </div>
  </div>
  <Overlay
    show={this.state.showColorPicker}
    placement="bottom"
    container={() => this.colorButtonRef.current}
  >
    <MiniColorPicker
      darkMode={ui.darkMode}
      chooseColor={this.chooseCardColor}
      el={this.colorButtonRef}
      close={() => this.setState({ showColorPicker: false })}
    />
  </Overlay>
</div>
```

Quiescent:
Good news! :)

I tested the current version to see whether I could spot what was going wrong and found it working(?!)
Is there something that I'm missing?

https://user-images.githubusercontent.com/1457336/123222412-360c6480-d4d0-11eb-91d9-0854b7fda601.mp4

(This was on MacOS).

cameronsutter:
@Quiescent when you say the current version, do you mean the way it is in this PR?

Quiescent:
> @Quiescent when you say the current version, do you mean the way it is in this PR?

Yes.

Double checked this morning and it's broken on staging and works perfectly on this branch (it doesn't bump other controls around and puts the mini colour picker in the right place).

cameronsutter (approved):

PR #683, opened 2021-06-18
fix book title overflow

jeanawhou:
ticket link: https://app.ora.pm/p/320746?c=5269086
fix preview

https://user-images.githubusercontent.com/19387007/122599841-8be0a700-d0a1-11eb-8ffc-5c0b5f474c90.mp4


Quiescent (approved):
This fixes one problem and, unfortunately, introduces another.
Now that the title is taller when the display is narrow, we have much too rounded corners.
Since this fixes the worse of the two problems, I vote that we merge it and that you work on a solution for the border radius on a follow-up PR.

PR #685, opened 2021-06-29
remove tags from card, character, note, places when tag is deleted

jeanawhou:
ticket link: https://app.ora.pm/p/320746?c=5230103
Sorry about the low resolution

preview
https://user-images.githubusercontent.com/19387007/123823198-8dc32980-d92f-11eb-83eb-bba505b04475.mp4


lib/plottr_components/src/components/tag/TagView.js
```diff
@@ -50,6 +50,52 @@ const TagViewConnector = (connector) => {
     deleteTag = (e) => {
       e.stopPropagation()
       this.props.actions.deleteTag(this.props.tag.id)
+      this.removeTags(this.props.tag.id)
+    }
+
+    removeTags = (tagId) => {
```
Quiescent: This isn't the right place to remove those tags.
The DELETE_TAG action should be handled in the `cards`, `characters`, `notes` and `places` reducers, where we can remove tags from each of those pieces of state.
(In an ideal world, our components would have no logic in them except for highly dynamic/interesting UI display stuff).

Quiescent:
Good job on the thorough testing!
Let's just move the logic to the reducers and then we can get this merged :)

lib/plottr_components/src/components/tag/TagView.js
Quiescent: This is nearly it.  The next step is to only have one action to delete a tag.
That action can go into the tags actions file.
You only need to call that one action and the `DELETE_TAG` event with the tag's `id` will reach all the reducers.

cameronsutter:
You're on the right path. Like what Ed is saying, you shouldn't need to create any new actions. The `deleteTag` action in actions/tags.js already fires a DELETE_TAG event. So the only change needed should be to change the cards/characters/places/notes reducers to handle the DELETE_TAG event

cameronsutter (changes requested):
You're on the right path. Like what Ed is saying, you shouldn't need to create any new actions. The `deleteTag` action in actions/tags.js already fires a DELETE_TAG event. So the only change needed should be to change the cards/characters/places/notes reducers to handle the DELETE_TAG event

Quiescent (approved):
Looks good.  LGTM!

PR #694, opened 2021-07-05
created manual webview element to allowpopups

jeanawhou:
ticket: https://app.ora.pm/p/320746?c=5381531
WDYT @cameronsutter @Quiescent ?

preview
https://user-images.githubusercontent.com/19387007/124499052-f444bd80-ddef-11eb-8475-e679c08ad077.mp4

### DIFF
--- a/src/dashboard/components/help/HelpHome.js
+++ b/src/dashboard/components/help/HelpHome.js
@@ -24,6 +24,13 @@ export default function HelpHome(props) {
     }
   }
 
+  const RenderWebView = () => {
+    return React.createElement('webview', {
+      src: 'https://docs.plottr.com',
+      allowpopups: 'true',
+    })
+  }
+
   return (
     <div className="dashboard__help">
       <div style={{ flex: 0.16 }}>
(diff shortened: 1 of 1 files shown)


src/dashboard/components/help/HelpHome.js
```diff
@@ -24,6 +24,13 @@ export default function HelpHome(props) {
     }
   }
 
+  const RenderWebView = () => {
+    return React.createElement('webview', {
```
Quiescent, line 28: Isn't this the same as `<webview src="https://docs.plottr.com" allowpopups></webview>`?
(Where the critical change is that we allowed popups.)

Quiescent: If so, then I'd suggest we inline it, rather than creating a separate function for it.

jeanawhou: @Quiescent i cant make it work using inline like what's on the docs, i believe the custom prop is being ignored 😕

Quiescent: Aha.  Ok.  Pitty :/

Quiescent (approved):
I've tested that this works and it's fine on my Mac.  LGTM :)

PR #696, opened 2021-07-15
Remove the save as button from the template picker

Quiescent:
The only case that we showed this button doesn't really work because there's confusion about whether we'd save a timeline template or a project template.


cameronsutter (approved):
Yup, that's exactly what i meant

jeanawhou (approved):

PR #702, opened 2021-07-21
Create a book from a template

cameronsutter:
This allows a user to create a new book within a project using a template.

It gets pretty tricky with the beats when making it work with act structure. Is there a better way to copy a beat tree and give it different ids?


lib/pltr/v2/reducers/beats.js
```diff
@@ -58,188 +59,190 @@ const addNodeToState = (state, bookId, position, title, parentId) => {
   }
 }
 
-const beats =
-  (dataReparires) =>
-  (state = INITIAL_STATE, action) => {
-    const actionBookId = associateWithBroadestScope(action.bookId)
+const beats = (dataReparires) => (state = INITIAL_STATE, action) => {
```
Quiescent: I think that you ran an older version of prettier on this file.
The new version prefers those curried function argument groups over several lines.
(It failed on CI: https://github.com/cameronsutter/plottr_electron/pull/702/checks?check_run_id=3128366013)

lib/plottr_components/src/components/project/BookList.js
```diff
@@ -57,6 +60,74 @@ const BookListConnector = (connector) => {
       beatActions.addBeat(newBookId)
     }
 
+    addBookFromTemplate = (template) => {
```
Quiescent: This is very hairy and definitely breaks the abstraction.

There are two things that I think would improve the situation:

1. I think that we should add support for this in the reducers rather than the components.
I think that it would be fair (in terms of abstraction) to place the burden on the reducer to know the current structure of beats in a template and to handle adding a new book from that template.
So how about creating a new action `ADD_BOOK_FROM_TEMPLATE` which receives the template as the payload.
The reducers each have to pick off their bit of state from the template and make adjustments accordingly.
Then the components go back to having no idea that it's a beat tree and that detail is hidden away in `pltr`.

2. Clearly there's a lot of work to add a whole beat tree in terms of minting beat ids.
To solve this problem, how about finding the next id: https://github.com/cameronsutter/plottr_electron/blob/e5a65eeb6a9e571770f2ae6acdbe8244391a40f2/lib/pltr/v2/helpers/beats.js#L86 (it always produces 1 + maximum id across all beat trees) and then adding it to all the beats in the tree by reducing over the tree (https://github.com/cameronsutter/plottr_electron/blob/e5a65eeb6a9e571770f2ae6acdbe8244391a40f2/lib/pltr/v2/reducers/tree.js#L205).

Here's what that might look like (I'm writing this in the PR so please don't use it verbatim!) :)

In `reducers/tree.js` add a `parentId` parameter to the callback function in `reduce` (and add a note to state that the `reduce` function must process parents to leaves in-order.
```javascript
// Contract: reduce must process the nodes in the supplied tree, from the root node, down to the children.
// This is important to the consumers of this function who might want to process the tree, knowing, that all
// ancestors of the current node were already processed.
export const reduce = (idProp) => (tree, f, initialValue) => {
  function iter(acc, nodeId, parentId) {
    const next = nodeId === null ? acc : f(acc, findNode(tree, nodeId))
    return children(tree, nodeId).reduce((acc, entity) => {
      return iter(acc, entity[idProp], nodeId)
    }, next)
  }

  return iter(initialValue, null, null)
}
```

In `reducers/beats.js`
```javascript
case ADD_BOOK_FROM_TEMPLATE:
  const beats = action.template.beats
  const idMap = {}
  let nextId = nextId(state)
  const newBeats = tree.reduce('id')(beats, (newBeatTree, nextBeat, parentId) => {
    const newId = nextId
    idMap[nextBeat.id] = newId
    nextId++
    const newParentId = idMap[parentId] || null
    const newBeat = {
      ...clone(nextBeat),
      id: newId,
    }
    return tree.addNode('id')(newBeatTree, newParentId, newBeat)
  }, tree.newTree('id'))
  const newBookId = Object.keys(state).map(Number.parseInt).reduce((thisId, thatId) => Math.max(thisId, thatId)) + 1
  return {
    ...state,
    [newBookId]: newTree,
  }
```

lib/pltr/v2/reducers/beats.js
cameronsutter: i wondered why it tried to format it this way. I'm on the latest version of prettier though (according to my vscode at least)

Quiescent:
That’s strange.  I noticed that I would get differences in formatting when I switched between node/npm versions.
Is VScode picking up the right version of npm/node?  (On MacOS the PATH is different between running an application via the shell and otherwise so that might cause problems(?))

lib/pltr/v2/reducers/beats.js
cameronsutter: @Quiescent i just had to reload vscode and it did it right

lib/plottr_components/src/components/project/BookList.js
cameronsutter: @Quiescent so i'm trying to implement this and i think one of the wrinkles that makes me think we need to do this outside of reducers (even though that would be the best if we could) is how to use the idMap that gets created for both lines and beats in the cards reducer?

To explain that more: since we have to change the IDs of beats and lines, we also have to update those IDs in the cards that come from the template. As far as I can tell, there's no way for the cards reducer to know what those new ids are unless we do the whole thing outside reducers

cameronsutter: ok, never mind. After thinking it through some more, i got what you said about adding the next id to the current id and got it working. This is the comment at the root reducer:

```
// cards from the template need to know the new ids of lines and beats from the template
// the strategy here is to use the state's next id value + the template id's current value
// the card reducer will have access to the state's next id value so it will be able to determine the correct id
```

Quiescent: Cool.  You're right about needing something in the root reducer to tie this all together.  :+1:

I'll take a closer look through the PR now and check to see if I can find problems with the implementation details.

Aside: there are similar problems here as we ran into when we tried to add Plot Lines from templates.  We could use some of the logic here to implement that feature too.  We still haven't resolved what it *means* to add a template based on an act structure into an existing act structure.
Do we:
 - put the two structures side by side?
 - embed the one structure into the other?
And what happens when there are too many beats?  (With the act structure off, we added a row of cards as pseudo beat titles, but that wouldn't work in an act structure).

lib/plottr_components/src/components/project/Book.js
```diff
@@ -114,19 +137,26 @@ const BookConnector = (connector) => {
...
         return (
           <div className={cx('book-container', 'add', { darkmode: ui.darkMode })}>
-            <div className="book add" onClick={this.props.addBook}>
+            {this.renderTemplatePicker()}
+            <div className="book add">
               <div className="front">
                 <div className="cover add">
                   <div className="book-container__add">
-                    <h1>
+                    <h3 onClick={() => this.props.addBook()}>
```
Quiescent: I've seen this a couple of times and just wanted to point out that there is a difference in performance between: `onClick={this.props.addBook}` and `onClick={() => this.props.addBook()}`.
The latter creates a new function each time that the component is rendered.
That means that the component that we're passing that lambda to will be redrawn because of a change in props.

lib/plottr_components/src/components/timeline/AddLineRow.js
```diff
@@ -192,6 +192,7 @@ const AddLineRowConnector = (connector) => {
       })
     }
 
+    // TODO: refactor by putting this logic in reducers
```
Quiescent, line 195: Ah!  I see that you're already thinking about adding plot lines from templates :)

Another benefit of moving this all to the reducers is that the mobile guys won't have to re-implement it.

lib/pltr/v2/reducers/beats.js
```diff
@@ -17,12 +17,14 @@ import {
...
   SET_HIERARCHY_LEVELS,
   LOAD_BEATS,
+  ADD_BOOK_FROM_TEMPLATE,
 } from '../constants/ActionTypes'
 import { beat as defaultBeat } from '../store/initialState'
 import { newFileBeats } from '../store/newFileState'
 import { positionReset, nextPositionInBook, moveNextToSibling } from '../helpers/beats'
 import { associateWithBroadestScope } from '../helpers/lines'
-import { addNode, deleteNode, editNode, nodeParent, filter, newTree, findNode } from './tree'
+// import { addNode, deleteNode, editNode, nodeParent, filter, newTree, findNode, reduce } from './tree'
```
Quiescent: Why not delete the import?

lib/pltr/v2/reducers/beats.js
```diff
@@ -31,9 +33,9 @@ import { clone } from 'lodash'
 //  - bookId: Number,
 //  - "series": String literal,
 
-const add = addNode('id')
+const add = tree.addNode('id')
 
-const defaultBeats = add(newTree('id'), null, defaultBeat)
+const defaultBeats = add(tree.newTree('id'), null, defaultBeat)
```
Quiescent, line 38: We call `tree.newTree('id')` enough that I think it might be worth binding `const newTree = tree.newTree(`id`)` along side the `addNode` binding.

lib/pltr/v2/reducers/root.js
```diff
@@ -21,6 +27,18 @@ const root = (dataRepairers) => (state, action) => {
   const isSeries = action.type.includes('@@') ? false : isSeriesSelector(state)
   const mainReducer = unrepairedMainReducer(dataRepairers)
   switch (action.type) {
+    case ADD_BOOK_FROM_TEMPLATE:
+      // cards from the template need to know the new ids of lines and beats from the template
+      // the strategy here is to use the state's next id value + the template id's current value
+      // the card reducer will have access to the state's next id value so it will be able to determine the correct id
```
Quiescent, line 33: Great comment!  I was about to mention that we needed something like this when I found it! :) :+1:

Quiescent (approved):
This is a great implementation and it'll reduce friction between mobile and web/desktop re. templates.

The only things I commented on are knit-pickings really so they shouldn't hold up the merge.

lib/plottr_components/src/components/project/Book.js
cameronsutter: oh yup, i did know that. Must have missed this one. Thanks

lib/plottr_components/src/components/timeline/AddLineRow.js
cameronsutter: YES!

lib/pltr/v2/reducers/beats.js
cameronsutter: oops. Forgot to clean that up. Thanks

PR #704, opened 2021-08-02
implement new template format

cameronsutter:
THIS IS NOT a new format for saving/storing templates. This is just using a more consistent template format that Ryan is implementing where beats are essentially not part of the template.

### DIFF
--- a/lib/pltr/v2/reducers/__tests__/root.test.js
+++ b/lib/pltr/v2/reducers/__tests__/root.test.js
@@ -1,0 +1,56 @@
+import { ADD_LINES_FROM_TEMPLATE } from '../../constants/ActionTypes'
+import rootReducerWithoutRepairers from '../root'
+import { beatsByPosition } from '../../helpers/beats'
+
+const rootReducer = rootReducerWithoutRepairers({
+  normalizeRCEContent: (x) => x,
+})
+
+describe('rootReducer', () => {
+  describe('ADD_LINES_FROM_TEMPLATE', () => {
+    const enoughBeats = require('./fixtures/enough_beats.json')
+    const notEnoughBeats = require('./fixtures/not_enough_beats.json')
+    const templateData = require('./fixtures/7_point_template.json')
+    const action = { type: ADD_LINES_FROM_TEMPLATE, templateData }
+    describe('enough beats', () => {
+      const result = rootReducer(enoughBeats, action)
+      const beatsAfter = beatsByPosition(() => true)(result.beats['1']).map(({ id }) => id)
+      it('doesnt add more beats', () => {
+        const beatsBefore = beatsByPosition(() => true)(enoughBeats.beats['1']).map(({ id }) => id)
+        expect(beatsBefore.length).toEqual(beatsAfter.length)
+      })
+      it('adds one line', () => {
+        expect(result.lines.length).toEqual(enoughBeats.lines.length + 1)
+      })
+      it('gives cards the right ids', () => {
+        // choose a random postion: 2
+        // there should be a card there
+        const beatId = beatsAfter[2]
+        const card = result.cards.find((c) => c.beatId == beatId)
+        expect(card).not.toEqual(undefined)
+        expect(card.title).toEqual('Pinch 1')
+      })
+    })
+    describe('not enough beats', () => {
+      const result = rootReducer(notEnoughBeats, action)
+      const beatsAfter = beatsByPosition(() => true)(result.beats['1']).map(({ id }) => id)
+      it('adds more beats', () => {
+        const beatsBefore = beatsByPosition(() => true)(notEnoughBeats.beats['1']).map(
+          ({ id }) => id
+        )
+        expect(beatsBefore.length).not.toEqual(beatsAfter.length)
+      })
+      it('adds one line', () => {
+        expect(result.lines.length).toEqual(notEnoughBeats.lines.length + 1)
+      })
+      it('gives cards the right ids', () => {
+        // choose a random postion (after the last beat in the file): 2
+        // there should be a card there
+        const beatId = beatsAfter[2]
+        const card = result.cards.find((c) => c.beatId == beatId)
+        expect(card).not.toEqual(undefined)
+        expect(card.title).toEqual('Pinch 1')
+      })
+    })
+  })
+})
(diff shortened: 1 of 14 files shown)


lib/pltr/v2/reducers/cards.js
```diff
@@ -75,8 +75,24 @@ const cards =
...
+            c.beatId = action.cardToBeatIdMap[c.id]
+          } else {
+            // TODO:
+            // think through this some more
+            // c.beatId = action.nextBeatId + c.beatId
+          }
+          c.lineId = action.nextLineId + c.lineId // give it the correct lineId
+          c.id = c.id + action.nextCardId // give it a new id
+          c.fromTemplateId = action.templateData.id
+          return c
```
Quiescent: There might be a subtle bug here.
This modifies the cards on the template.
If you add the same template to two separate timelines, then I think you'll end up modifying the cards on both timelines(!)
Probably best to clone the card to be safe.

lib/pltr/v2/reducers/lines.js
```diff
@@ -73,16 +73,18 @@ const lines =
...
+        ).length
+        const newLines = action.templateData.lines.map((l, index) => {
           l.id = l.id + action.nextLineId // give it a new id
-          l.bookId = action.newBookId // add it to the new book
+          l.bookId = action.newBookId || action.bookId // add it to the new/current book
           l.fromTemplateId = action.templateData.id
+          if (!l.color || l.color == nextColor(0)) {
+            l.color = nextColor(linesInBook_ + index)
+          }
           return l
```
Quiescent: There's a similar problem with modifying objects here.

lib/pltr/v2/reducers/root.js
```diff
@@ -39,6 +43,63 @@ const root = (dataRepairers) => (state, action) => {
...
+      // ALSO:
+      // TODO:
+      // ignore beats unless there are > 1 hierarchyLevels
+      // add cards to leaf nodes (beats) unless there are > 1 hierarchyLevels
+      const bookId = state.ui.currentTimeline
+      let nextIdForBeats = nextBeatId(state.beats)
+      let beatTree = cloneDeep(state.beats[bookId])
+      let createdNewBeats = false
+      // make a card -> beatId mapping (beatId is from existing beats … augmented with new ones)
+      const beatPositionMap = Object.keys(beatTree.index).reduce((acc, id) => {
```
Quiescent: I think we should use `tree.reduce` instead so that the root reducer is insulated from the implementation of beats.

lib/pltr/v2/reducers/root.js
```diff
@@ -39,6 +43,63 @@ const root = (dataRepairers) => (state, action) => {
...
+        acc[beatTree.index[id].position] = parseInt(id)
+        return acc
+      }, {})
+      const cardToBeatIdMap = action.templateData.cards.reduce((acc, card) => {
+        const beat = action.templateData.beats['1'].index[card.beatId]
```
Quiescent: Using `tree` this is `tree.findNode(action.templateData.beats['1'], card.beatId)` as apposed to `.index[card.beatId]`.

lib/pltr/v2/reducers/root.js
Quiescent: On second thought: we have a helper function that flattens a tree in sorted order already:
`helpers/beats.js:beatsByPosition`.
This function insulates the reducer from another detail: the fact that beats have a `position` and simplifies this function: it would become something like:
```
beatsByPosition(() => true)(beatTree).map(({ id }) => id)
```

Quiescent: It doesn't make much sense for deeper hierarchies because you'd overlay the cards over the hierarchy regardless of level.

Quiescent (commented):
I like the approach, and the implementation has just a few niggles that I pointed out in the review.

I think that this is complicated enough that it warrants a test -- especially since three projects now rely on it and can change it on their own.
It would be useful if a contract enforced the expected behaviour so that each project doesn't end up accidentally breaking it.

One other thing that I think we ought to remember is that the migrator is probably not going to do something sensible with a template that lacks beat ids.
Is the idea to create templates as we have and to attach the `ignoreBeats` property in the action creator?

cameronsutter:
> I think that this is complicated enough that it warrants a test -- especially since three projects now rely on it and can change it on their own.

👍

> It would be useful if a contract enforced the expected behaviour so that each project doesn't end up accidentally breaking it.

What form do you imagine this contract being in?

> One other thing that I think we ought to remember is that the migrator is probably not going to do something sensible with a template that lacks beat ids.
> Is the idea to create templates as we have and to attach the `ignoreBeats` property in the action creator?

Yes. We still need templates with beats because when a user creates a new project or book with a template, we still use them.  So templates don't need to change actually, but in this specific case of adding a new plotline from the template, we ignore them.
I realized last night that there's that distinction between those two use cases that I was confusing a bit in my comments/code. So i'll go through and make that more obvious now that I've got it clear

Quiescent:
> > It would be useful if a contract enforced the expected behaviour so that each project doesn't end up accidentally breaking it.
> 
> What form do you imagine this contract being in?

This was meant to be part of the same idea relating to tests: the tests would be the contract.

> Yes. We still need templates with beats because when a user creates a new project or book with a template, we still use them. So templates don't need to change actually, but in this specific case of adding a new plotline from the template, we ignore them.
> I realized last night that there's that distinction between those two use cases that I was confusing a bit in my comments/code. So i'll go through and make that more obvious now that I've got it clear

:+1:

lib/pltr/v2/reducers/__tests__/root.test.js
```diff
@@ -0,0 +1,56 @@
...
+    const enoughBeats = require('./fixtures/enough_beats.json')
+    const notEnoughBeats = require('./fixtures/not_enough_beats.json')
+    const templateData = require('./fixtures/7_point_template.json')
+    const action = { type: ADD_LINES_FROM_TEMPLATE, templateData }
+    describe('enough beats', () => {
+      const result = rootReducer(enoughBeats, action)
+      const beatsAfter = beatsByPosition(() => true)(result.beats['1']).map(({ id }) => id)
+      it('doesnt add more beats', () => {
+        const beatsBefore = beatsByPosition(() => true)(enoughBeats.beats['1']).map(({ id }) => id)
+        expect(beatsBefore.length).toEqual(beatsAfter.length)
```
Quiescent, line 20: BTW, `expect` has a better way of checking array equality: https://jestjs.io/docs/expect#expectarraycontainingarray
(Not that I'm asking you to change the test because this provides good enough validation.)

Quiescent (approved):
Looks great!  LGTM (but maybe after cutting the release?)

PR #705, opened 2021-08-03
Bunch undoable actions together

Quiescent:
# Motivation
Undo is inconsistent across the app.
This is because Plottr undoes changes programmatically (using `redux-undo`) in almost every case, but it doesn't use it for RCEs.
This PR makes undo consistent.

# Details
Changes are bunched into undo "transactions" if:
 - the action type and the action keys are the same,
 - there are fewer than twenty edits in a transaction,
 - the same element was edited (i.e. if you switch between RCEs then a transaction boundary is introduced), and
 - edits occur within a ten-second window.

# Challenge
It's not enough to simply change the `value` or `text` of an RCE because the RCE is often left with an invalid selection.
i.e. a selection that points to a position in the `text` that doesn't exist.

# Solution
All editors current `selection`s are stored in a new root key for the file called `editors`.
A unique path per editor is minted to store that editor's `selection`.
If an editor is supplied with a selection, it'll use programmatic undo, otherwise it'll default to using the browser (Electron's) built-in undo.

# Extra Changes
Cards, Notes, Characters and Places are more uniform ITO how attributes are edited:
 - more logic moved to reducers,
 - new action for editing template attributes,
 - all three use the `EditAttribute` component for custom and template attributes.
 
 # Future Work
 Because we now store the `selection` for all editors, it's possible to remember where the user left off in every editor.
 I've left this for another PR because it's not as simple as setting the `selection`.


src/app/index.js
```diff
@@ -209,22 +209,8 @@ ipcRenderer.on('move-from-temp', () => {
...
-      editor.undo()
-    }
-  } else {
-    // custom undo function
-    store.dispatch(ActionCreators.undo())
-  }
+  store.dispatch(ActionCreators.undo())
 })
 
 ipcRenderer.on('redo', (event) => {
```
Quiescent: I missed this one.

problems-with-current-implementation.org
```diff
@@ -0,0 +1,17 @@
+ - [X] I need to support each type of RCE edit action at the moment.
```
Quiescent: I should have deleted this.

Quiescent:
I need to deal with the one char remaining bug.

Also, looking at the recording again, it seems that undo isn't batching edits in some of the text inputs (as apposed to RCEs).

Quiescent:
> I need to deal with the one char remaining bug.
> 
> Also, looking at the recording again, it seems that undo isn't batching edits in some of the text inputs (as apposed to RCEs).

I fixed both of these.

cameronsutter (approved):
LGTM

This was a large change, so even though I think i get it, I'm sure there will be some things that I missed

PR #706, opened 2021-08-03
Feature/drag drop stack cards

jeanawhou:
ticket: https://app.ora.pm/p/320746?c=6049791

preview:

https://user-images.githubusercontent.com/19387007/128056144-9db1f0ee-7dc6-41a9-8790-a234509d1645.mp4

### DIFF
--- a/lib/plottr_components/src/components/timeline/BlankCard.js
+++ b/lib/plottr_components/src/components/timeline/BlankCard.js
@@ -57,8 +57,11 @@ const BlankCardConnector = (connector) => {
       if (!droppedData.cardId) return
 
       const { beatId, lineId } = this.props
-
-      this.props.actions.reorderCardsWithinLine(beatId, lineId, [droppedData.cardId])
+      if (droppedData.cardId.length > 1) {
+        this.props.actions.reorderCardsWithinLine(beatId, lineId, [...droppedData.cardId])
+      } else {
+        this.props.actions.reorderCardsWithinLine(beatId, lineId, [droppedData.cardId])
+      }
     }
 
     saveCreate = () => {
--- a/lib/plottr_components/src/components/timeline/BlankCard.js
+++ b/lib/plottr_components/src/components/timeline/BlankCard.js
@@ -54,14 +54,15 @@ const BlankCardConnector = (connector) => {
 
       const json = e.dataTransfer.getData('text/json')
       const droppedData = JSON.parse(json)
-      if (!droppedData.cardId) return
 
       const { beatId, lineId } = this.props
-      if (droppedData.cardId.length > 1) {
-        this.props.actions.reorderCardsWithinLine(beatId, lineId, [...droppedData.cardId])
-      } else {
+      if (droppedData.cardIds) {
+        this.props.actions.reorderCardsWithinLine(beatId, lineId, [...droppedData.cardIds])
+      } else if (droppedData.cardId) {
         this.props.actions.reorderCardsWithinLine(beatId, lineId, [droppedData.cardId])
       }
+
+      return
     }
 
     saveCreate = () => {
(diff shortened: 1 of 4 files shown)


lib/plottr_components/src/components/timeline/CardCell.js
```diff
@@ -17,7 +18,89 @@ const CardCellConnector = (connector) => {
...
+      e.stopPropagation()
+      e.preventDefault()
+
+      if (this.state.dragging) return
+      if (!allowDrop) return
+      this.setState({ inDropZone: false, dropDepth: 0 })
+
+      const json = e.dataTransfer.getData('text/json')
+      const droppedData = JSON.parse(json)
+      if (!droppedData.cardId) return
```
Quiescent: What if it's a stack of cards?
Wouldn't `droppedData.cardIds` be bound then?

lib/plottr_components/src/components/timeline/CardCell.js
```diff
@@ -17,7 +18,89 @@ const CardCellConnector = (connector) => {
...
+        hovering: false, // forces a rerender to determine where to place the overlay
+      }
+    }
+
+    handleDragStart = (e) => {
+      this.setState({ dragging: true })
+      e.dataTransfer.effectAllowed = 'move'
+      const cardIds = this.props.cards.map((card) => card.id)
+
+      if (!Array.isArray(cardIds) && cardIds.length <= 1) return
```
Quiescent: Shouldn't this be or (`||`)?
If it's not an array, it surely won't have a length.

Or did you perhaps mean to check whether it *is* an array and has a length less than or equal to 1?

lib/plottr_components/src/components/timeline/BlankCard.js
```diff
@@ -54,11 +54,15 @@ const BlankCardConnector = (connector) => {
 
       const json = e.dataTransfer.getData('text/json')
       const droppedData = JSON.parse(json)
-      if (!droppedData.cardId) return
 
       const { beatId, lineId } = this.props
+      if (droppedData.cardIds) {
```
Quiescent, line 59: Is this the only other place that the user can drop cards?
Thinking about it, it makes sense that this would be the only place, but it's probably best to be sure.

lib/plottr_components/src/components/timeline/CardCell.js
Quiescent: Did you drop a stack of cards onto another stack of cards in your testing?

Quiescent (commented):
Looking great!  I think that the problems I found are really minor.
Please ping me when you're ready for a follow-up review :)

lib/plottr_components/src/components/timeline/CardCell.js
jeanawhou: Yeah i meant, the second one 😅 but yeah `||` is right

lib/plottr_components/src/components/timeline/CardCell.js
jeanawhou: Yeah i have tried but its not allowed yet. Im sure we i should also allow stacked cards to be droppable to another stack of cards 😅 wdyt?

Quiescent: I think it should be allowed.  It seems unexpected to forbid that but allow one card to be dropped onto a stack. :)

Quiescent:
There seems to be still a problem re-rendering the "collapsed" state of lines.
See this recording I took:

https://user-images.githubusercontent.com/1457336/128472162-201913f2-f52d-4db8-a142-071136ca1659.mp4

Quiescent:
Here's the last bug I've found (you can't drop a stack onto a single card):

https://user-images.githubusercontent.com/1457336/128831740-aae3419a-e584-4fdd-b3e6-35d6dcab3fbd.mp4

jeanawhou:
> Here's the last bug I've found (you can't drop a stack onto a single card):

Sorry i should have asked for this one since i already noticed it 😅

Quiescent (approved):
Looks good.  LGTM! :)

PR #708, opened 2021-08-10
Integrate Dashboard Into the Same Window as the App

Quiescent:
# Motivation
The design direction of Plottr aligns better with one window per document.
This also works better for a web page where multiple tabs are unwieldy.

# Changes
This PR makes big changes to the side effects in the application.
Most interactions with the file system are now in the main process.

## List of Changes
 - Create a popover on the navbar to open the dashboard.
 - Refactor the Navigation to a component.
 - Put the dashboard in a modal accessible from the navbar.
 - Various style tweaks to make dashboard classes work while embedded.
 - Move dashboard components to the components library.
 - Extract platform-specific dependencies into `connector`.
 - Remove the dashboard entry point from Webpack and the source tree.
 - Initialise `format-message` early to eliminate uninitialised logs.
 - Open the dashboard automatically the first file that's opened and then not again.
 - Tweak modal designs for a more full-screen display of the dashboard.


Quiescent:
I went through this with a fine-toothed comb to try and pick up mistakes proactively.
The only thing that I found that I don't think was worth fixing yet is that the backup options don't sync immediately to other windows with the settings open (all other settings seem to).

PR #709, opened 2021-08-12
fix missing words from lists with font style combination

jeanawhou:
@cameronsutter @Quiescent i have no luck fixing the numbered list 😅 then i noticed there's a separate ticket for it so this fix is only for the missing words when exporting the lists to word with combination of font styles.

Ticket: https://app.ora.pm/p/320746?c=6058749
preview:
https://user-images.githubusercontent.com/19387007/129218620-ed0f0e7b-895d-4231-a7ae-8531ce60f255.mp4

### DIFF
--- a/src/common/slate_serializers/to_word.js
+++ b/src/common/slate_serializers/to_word.js
@@ -51,15 +51,18 @@ const serialize = (nodes, doc) => {
 
     switch (n.type) {
       case 'bulleted-list':
-        return children.map((li) => {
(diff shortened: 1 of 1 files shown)


Quiescent (approved):
Looks good, LGTM!

Quiescent:
(To be sure, I also tested whether we're still exporting correctly when the bullets are in a quote and that worked too :) )

PR #711, opened 2021-08-17
Move backup to main and only backup when idle for a second

Quiescent:
# Motivation
In huge files, the application hangs when we make many consecutive edits to the file.
Many edits in quick succession happen when we edit an RCE or an input field whose value is controlled by the store.

# Solution
## Move Backup to the Main Process
This means that the renderers don't need to wait for IO operations regarding the file.

## Wait for Idle Before Saving
Even without waiting for IO, the renderers still take time to auto-save because they need to serialise the file before sending it to the main process.
It's not catastrophic for a second or so's worth of work to be lost.
So instead, in this PR, Plottr waits for a second before saving.
If another edit happens in that window, we wait for a second from that point in time.
If we bump back saving 100 times, then we force a save because that's too many edits not to save.


cameronsutter (approved):
LGTM

PR #712, opened 2021-08-17
Atomic swap save file

Quiescent:
# Motivation
When Plottr crashes or slows down, we sometimes lose part or all of the file on-disk.
This bug is a huge problem for users, who might lose work and restore their backups files.
Unfortunately, I haven't been able to replicate this problem reliably.
This solution should solve the problem because it behaves atomically (i.e. all or nothing) when saving the file.

# Solution
In this PR, we save the file to a temporary name synchronously, and when the file system reports that the new file exists, we swap the old file for the temporary file.
This atomic-swap-like solution should be much more robust.


cameronsutter (approved):
LGTM

PR #713, opened 2021-08-17
Feature/droppable empty outline chapters

jeanawhou:
ticket: https://app.ora.pm/p/320746?c=6027413
preview:

https://user-images.githubusercontent.com/19387007/129776841-27ba6ec6-52b1-4489-b201-d1ba7ec667a5.mp4

### DIFF
--- a/lib/plottr_components/src/components/outline/BeatView.js
+++ b/lib/plottr_components/src/components/outline/BeatView.js
@@ -140,7 +140,7 @@ const BeatViewConnector = (connector) => {
       })
     }
 
-    renderDropZone() {
+    renderDropZone = () => {
       if (!this.state.inDropZone) return
       if (!this.props.cards[0].isEmpty) return
 
--- a/lib/plottr_components/src/components/outline/BeatView.js
+++ b/lib/plottr_components/src/components/outline/BeatView.js
@@ -6,6 +6,7 @@ import { t as i18n } from 'plottr_locales'
 import UnconnectedCardView from './CardView'
 import cx from 'classnames'
 import { helpers } from 'pltr/v2'
+import { FaCircle } from 'react-icons/fa'
 
 const {
   card: { sortCardsInBeat },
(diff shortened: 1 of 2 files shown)


Quiescent:
At 17s, it looks like the drop zone appears above chapter 10 instead of beneath it.
It feels a little off to me.
How difficult would it be to put the drop zone beneath the chapter?

lib/plottr_components/src/components/outline/BeatView.js
```diff
@@ -88,7 +89,9 @@ const BeatViewConnector = (connector) => {
     }
 
     renderManualSort() {
-      if (this.props.beat.autoOutlineSort) return null
+      const { cards, beat } = this.props
+      if (cards[0].isEmpty || cards.length === 1) return
```
Quiescent, line 98: I don't think that there's an `isEmpty` attribute on cards(?)

lib/plottr_components/src/components/outline/BeatView.js
```diff
@@ -128,8 +176,14 @@ const BeatViewConnector = (connector) => {
...
         >
-          <div>
+          <div
+            onDragLeave={this.handleDragLeave}
+            onDragEnter={this.handleDragEnter}
+            onDragOver={this.handleDragOver}
+            onDrop={this.handleDrop}
+          >
             <h3 id={`beat-${beat.id}`} className={klasses}>
+              {this.renderDropZone()}
```
Quiescent: I think that if you put this after the `beatTitle` then it should render beneath the title when you drag and drop.

lib/plottr_components/src/components/outline/BeatView.js
Quiescent: Were you thinking of the `lodash` function `isEmpty`?

lib/plottr_components/src/components/outline/BeatView.js
jeanawhou: Yeah it looks more consistent. Thank you @Quiescent

lib/plottr_components/src/components/outline/BeatView.js
jeanawhou: @Quiescent the `isEmpty` attribute is only present if the chapter is empty. I added the check because multiple dropzone is showing when dragging a card. The dropzone for CardView and BeatView both are visible.
image

Quiescent (approved):
Looks good.  LGTM! :)

PR #719, opened 2021-08-24
fix export plotline filter

jeanawhou:
Ticket: https://app.ora.pm/p/320746?c=6027419
I noticed the issue pointed out on the card only in scriv so the video attached only shows for the scriv 😅 

preview

https://user-images.githubusercontent.com/19387007/130655019-d79c4c0e-98ee-4dd5-8175-d6f74ed06821.mp4


Quiescent (approved):
Looks good.  LGTM

PR #729, opened 2021-09-15
onboarding flow - semi skeleton

cameronsutter:


lib/plottr_components/src/components/FirebaseLogin.js
```diff
@@ -0,0 +1,70 @@
...
+
+    useEffect(() => {
+      if (firebaseLoginComponentRef) {
+        const ui = firebaseUI()
+        startUI(ui, '#firebase_login_root')
+      }
+    }, [firebaseLoginComponentRef])
+
+    useEffect(() => {
+      onSessionChange((user) => {
```
Quiescent: `onSessionChange` returns a function that you can call to deregister the listener.
(I need to update the README.)

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep0.js
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <OnboardingStep>
+      <StepHeader>
+        <h2>{t('You got Plottr Pro 🎉')}</h2>
+        <p>{t("Let's get you onboarded")}</p>
+      </StepHeader>
+      <StepBody>
+        <div className="onboarding__well">
+          <p className="accented-text">
+            {t("This will be quick and painless. There's only 3 steps:")}
```
Quiescent: s/There's/There are/ ?

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep0.js
```diff
@@ -0,0 +1,45 @@
...
+            <li>{t('Sign in')}</li>
+            <li>{t('Upload projects and templates to the cloud')}</li>
+            <li>{t('A few easy settings')}</li>
+          </ol>
+        </div>
+      </StepBody>
+      <StepFooter>
+        <OnboardingButtonBar>
+          <Button bsSize="large" onClick={cancel}>
+            {t('Not Now')}
```
Quiescent: How does the user get back here?
If the pro license is activated, won't it skip the account dialog?

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep1.js
```diff
@@ -0,0 +1,44 @@
...
+    platform: { useSettingsInfo },
+  } = connector
+
+  const FirebaseLogin = UnconnectedFirebaseLogin(connector)
+
+  const ProStep1 = ({ nextStep }) => {
+    const [settings] = useSettingsInfo()
+    useEffect(() => {
+      if (settings.user.id) nextStep()
+    }, [settings])
```
Quiescent, line 32: Is `settings` enough here?
I think that the object is mutable, so the hook won't see the setting change.

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep1.js
```diff
@@ -0,0 +1,44 @@
...
+
+    const checkUserId = (uid) => {
+      if (uid) nextStep()
+    }
+
+    const showFrb = settings.user && !settings.user.id
```
Quiescent: I prefer connecting components to redux and using `client.userId` from there because all of the places that we use these storage hooks have odd race conditions that I've had to debug.

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep2.js
```diff
@@ -0,0 +1,104 @@
...
+import PropTypes from 'react-proptypes'
+import { t } from 'plottr_locales'
+import OnboardingStep from '../../../onboarding/OnboardingStep'
+import { StepBody, StepHeader } from '../../../onboarding/Step'
+import UnconnectedUploading from './Uploading'
+
+// is there a way to know that this step is done?
+
+const isPlottrCloudFile = (filePath) => filePath && filePath.startsWith('plottr://')
+// needs to be declared here otherwise it will cause an infinite loop since we'd be passing in a new [] every render
```
Quiescent: You could use a ref instead.
`const initialFrbFiles = useRef([])`

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep2.js
```diff
@@ -0,0 +1,104 @@
+import React, { useState, useEffect } from 'react'
+import PropTypes from 'react-proptypes'
+import { t } from 'plottr_locales'
+import OnboardingStep from '../../../onboarding/OnboardingStep'
+import { StepBody, StepHeader } from '../../../onboarding/Step'
+import UnconnectedUploading from './Uploading'
+
+// is there a way to know that this step is done?
```
Quiescent: Migration is a per-machine step, so we could use settings (or something else that's persisted to the HDD).

We can also go through the file list and find the files that aren't any cloud files, but that doesn't mean that the user skipped this step.

A combination of the two might be appropriate.
Track migrations using state persisted to the HDD and notify them that they could migrate more to the cloud if we find more files there.

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep2.js
```diff
@@ -0,0 +1,104 @@
...
+  const ProStep2 = ({ nextStep }) => {
+    const [sortedIds, filesById] = useSortedKnownFiles('', initialFrbFiles)
+    const [templateInfo] = useCustomTemplatesInfo()
+    const [nothing, setNothing] = useState(false)
+    const [view, setView] = useState('choice')
+    const [templates, setTemplates] = useState(null)
+    const [projects, setProjects] = useState(null)
+
+    useEffect(() => {
+      // TODO: figure out what to do with templates/projects already in firebase
```
Quiescent, line 33: Surely nothing?
What else would we do?

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep2.js
```diff
@@ -0,0 +1,104 @@
+import React, { useState, useEffect } from 'react'
+import PropTypes from 'react-proptypes'
+import { t } from 'plottr_locales'
+import OnboardingStep from '../../../onboarding/OnboardingStep'
+import { StepBody, StepHeader } from '../../../onboarding/Step'
+import UnconnectedUploading from './Uploading'
+
+// is there a way to know that this step is done?
+
+const isPlottrCloudFile = (filePath) => filePath && filePath.startsWith('plottr://')
```
Quiescent, line 10: We should probably think about abstracting this out at some point.
I think it's repeated about 4 times in the code base now :/

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep2.js
```diff
@@ -0,0 +1,104 @@
...
+        setNothing(true)
+      } else {
+        setNothing(false)
+      }
+    }, [sortedIds, templateInfo])
+
+    useEffect(() => {
+      // there's nothing to upload
+      if (templateInfo && sortedIds && nothing) nextStep()
+    }, [nothing])
```
Quiescent: Don't you want `templateInfo` and `sortedIds` in the effect list?

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep2.js
```diff
@@ -0,0 +1,104 @@
...
+
+    if (!templates && !projects && !nothing) {
```
Quiescent: `!nothing` feels confusing and might cause a glitch in rendering.
If we initialised it to `null` then we'd have the semantics down very precisely to "when stuff hasn't yet loaded".

lib/plottr_components/src/components/dashboard/account/proOnboarding/Uploading.js
```diff
@@ -0,0 +1,143 @@
...
+const typeName = {
+  project: t('Project'),
+  template: t('Template'),
+}
+
+const UploadingConnector = (connector) => {
+  const {
+    platform: {
+      file: { basename, doesFileExist, readFileSync },
+      firebase: { saveCustomTemplate },
```
Quiescent: ping

lib/plottr_components/src/components/dashboard/account/proOnboarding/Uploading.js
```diff
@@ -0,0 +1,143 @@
...
+      //     uploadProject(obj)
+      //   } else {
+      //     // upload template
+      //     // saveCustomTemplate(settings.user.id, obj.data)
+      //   }
+      // })
+      // // done! Go to the next step
+      // setCurrentObj(t('Done!'))
+      // setTimeout(nextStep, 1500)
+    }, [settings])
```
Quiescent, line 72: Do you really want to re-run the effect at all?  I think that would duplicate uploading tasks.

lib/plottr_components/src/components/dashboard/account/proOnboarding/Uploading.js
```diff
@@ -0,0 +1,143 @@
...
+
+    const uploadProject = (projObj) => {
```
Quiescent, line 74: Now that I'm reading this PR, I have a better idea for uploading the file.
Instead of loading the file and relying on sync, we should use: `overwriteAllKeys(fileId, clientId, state)`
Migration is probably still a good idea.

lib/plottr_components/src/components/dashboard/account/proOnboarding/ProStep0.js
cameronsutter: yeah that's something else to handle … what happens when they start with Pro from the beginning? Thanks for bringing this up

PR #730, opened 2021-09-17
enable finish editing onclickedOutside and esc key

jeanawhou:
@cameronsutter @Quiescent i enabled the keyboard and click events on my screen recorder so the changes ive made will be visible from the preview :sweat_smile: 
What do you think guys?

Ticket: https://app.ora.pm/p/320746?c=6026975
Preview: 

https://user-images.githubusercontent.com/19387007/133806404-c3530ae8-605e-4653-8ff0-fdf02802b608.mp4

### DIFF
--- a/lib/plottr_components/src/components/outline/BeatView.js
+++ b/lib/plottr_components/src/components/outline/BeatView.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { Component, createRef } from 'react'
 import PropTypes from 'react-proptypes'
 import { Glyphicon } from 'react-bootstrap'
 import { Waypoint } from 'react-waypoint'
@@ -20,7 +20,9 @@ const BeatViewConnector = (connector) => {
   const CardView = UnconnectedCardView(connector)
 
   class BeatView extends Component {
-    state = { sortedCards: [], inDropZone: false, dropDepth: 0 }
+    state = { sortedCards: [], inDropZone: false, dropDepth: 0, clickedOutside: false }
+    beatRef = createRef()
+    titleRef = createRef()
 
     static getDerivedStateFromProps(nextProps, nextState) {
       const { beat, cards, lines } = nextProps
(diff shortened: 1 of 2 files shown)


lib/plottr_components/src/components/outline/CardView.js
```diff
@@ -216,6 +216,12 @@ const CardViewConnector = (connector) => {
       return <div className="divider" />
     }
 
+    UNSAFE_componentWillReceiveProps(nextProps, nextState) {
```
Quiescent: Could we please try to avoid deprecated lifecycle hooks?
We could achieve the same effect if we used `componentDidUpdate` instead.
Alternatively, I think it would be worth the effort to refactor this to a function component and use a hook.

lib/plottr_components/src/components/outline/BeatView.js
```diff
@@ -158,6 +166,30 @@ const BeatViewConnector = (connector) => {
...
+    componentDidMount() {
+      document.addEventListener('click', this.handleClickOutside, true)
+    }
+
+    componentWillUnmount() {
+      document.removeEventListener('click', this.handleClickOutside, true)
+    }
+
+    handleClickOutside = (e) => {
+      if (this.titleRef.current && this.titleRef.current.contains(e.target)) {
```
Quiescent, line 178: I'm afraid I have to disagree with the timeout here because it adds extra complexity and re-renders.
What if we initialised a piece of state in the `CardView` to a timestamp.
Then you can set `clickedOutside` to a timestamp in `BeatView` whenever the user clicks outside.
The `CardView` can compare the timestamp to its own.
If the timestamp is different, it calls `saveEdit`.

One other problem here is that the if statements should be combined because the body of the first if statement is identical to the second.

jeanawhou: @Quiescent I removed the setTimeout and it didn't break anything, but im not completely sure if that has another downside though. What do you think Ed?

Quiescent: Ah!  The callback argument of `setState`.
Yes.  That'll do.

Quiescent: I see that you didn't collapse the if statements.
Because the whole block is so small, I'm not going to block merging because of it, but I'm going to point out that the body of the first condition and the else are the same.

Quiescent (approved):
LGTM! :)

PR #731, opened 2021-09-26
fix remove license bug

jeanawhou:
@Quiescent @cameronsutter no ticket for this bug, i just noticed this after i pulled the latest.
The bug can be reproduce when you click on the `Remove License` button from the Account View. 
It will throw an error it will throw an error like this 
image

fix preview:

https://user-images.githubusercontent.com/19387007/134818353-568b67c8-afb8-4ec9-bef7-ea638a743cc9.mp4


Quiescent (approved):
Nice catch!  LGTM!

PR #738, opened 2021-11-23
Offline Mode

Quiescent:
# Motivation
We'd like users to be able to continue using Plottr (pro) while they're not connected to the Internet.

# Solution
This PR tracks the online status of the application and whether we're resuming as two flags in the `project` key of the Redux store.
Additionally, Plottr now backs up all files to a new folder in the Plottr app directory (`~/.config/plottr/offline`) with a file name derived from the name of the cloud file and sanitised to use for a path on an HDD.

## Offline Mode
When we detect that we've gone offline, we display a warning banner along the bottom of all windows.
All changes are made to a backup file stored in the Plottr app directory under a new folder named `offline`.

## Resuming
There are two cases of resuming online files that Plottr is cognisant of:
 1. We boot/open a file that we made offline changes to.
 2. We go from offline to online while editing a file.

### From Offline
When the app goes back online, we check whether there were changes to the offline file and the online file.
 - If there were changes to the offline file and the online file, we backup the offline file and load the cloud file.
 - If there were changes to the offline file and there weren't changes to the online file, we over-write the cloud file.
 - If there were changes to the online file and no changes to the offline file, we load the online file.
 - If there were no changes to the online file and there were no changes to the offline file, we load the cloud file.

### When Booting A File
When the app boots a file, it checks for an offline backup and:
 - If there were changes to the offline file and the online file, we backup the offline file and load the cloud file.


src/app/components/Resume.js
```diff
@@ -0,0 +1,150 @@
...
+                setResuming(false)
+                setOverwriting(false)
+                retryCount = 0
+              })
+            }
+          })
+        })
+      }
+      function handleError(error) {
+        logger.error('Error trying to resume online mode', error)
```
jeanawhou, line 110: @Quiescent im not sure if your already notice this one.

jeanawhou (commented):
The changes looks good to me 😅

lib/plottr_components/src/components/dashboard/files/RecentFiles.js
```diff
@@ -42,10 +44,40 @@ const renderPermission = (permission) => {
...
+  }, {})
+  const filteredFileIds = Object.keys(filesById).filter((id) => {
+    if (searchTerm && searchTerm.length > 1) {
+      const f = filesById[`${id}`]
+      return (f.fileName || f.path).toLowerCase().includes(searchTerm.toLowerCase())
+    } else {
+      return true
+    }
+  })
+  const sortedFileIds = sortBy(filteredFileIds, (id) => filesById[`${id}`].lastOpened).reverse()
```
cameronsutter: fyi I've changed this line in staging, and that makes me think this should be a function that we pull out into a helper file so we don't have to paste it here

cameronsutter (commented):
Just had one small suggestion. Otherwise looks good

src/app/components/Resume.js
Quiescent: Yup.  This is intentional :)

Quiescent: Thanks for pointing it out though.

lib/plottr_components/src/components/dashboard/files/RecentFiles.js
Quiescent: Thanks for pointing that out! I'll take a look at consolidating them.

PR #739, opened 2021-12-08
11 9 release

Quiescent:
- When computing pro info: check for a user before getting their token to check for beta and admin claims.
 - When computing pro info: request pro info from API when there is no `currentUser` (from Firebase lib), but there is an email address that we can check with.
 - Delete some unused dependencies from the `VerifyPro` component.
 - During onboarding: allow the user to continue onboarding without uploading anything.
 - Move `showFrb` to an effect during onboarding (not sure why).
 - In the dashboard, don't check for an active license when we already know that we have one.
 - Add a slight delay in atomic save swap to accommodate OS disk contention with other programs that manage files (like Dropbox).
 - Fix the logic that decides what to show from `Navigation` when the app is started.
 - When checking for pro, account for `currentUser` being undefined.
 - Force Webpack to import and use the root `docx` everywhere, including sub-packages.


Quiescent:
Closed by squashing the branch and merging locally before pushing to staging.

PR #746, opened 2022-01-24
Upgrade electron

Quiescent:
# Motivation
Electron is the core of our application.  So with every upgrade, we get performance and security improvements (from the newer versions of Chromium that it bundles), better support for new libraries, and we minimize the risk that we'll be left dead in the water when a critical vulnerability is discovered.

# Considerations
## Less Node on Renderers
Every new release of Electron seems to eliminate some features of Electron that are both convenient and make it more like a desktop app.
They've gone this way because more and more developers are embedding web components and content into Electron applications.
It's a problem to have a web environment that's not sand-boxed and might have access to file-system and OS features, so the folks over at Electron decided to step in and start culling these features before they're too widely used.
This upgrade used a combination of restoring features (because, in our case, it's safe to do so) and re-implementing others.

In this release, we needed to:
 - add the `@electron/remote` library and configure it from the main process (`remote` is no longer bundled with `electron`),
 - use `import` syntax instead of node's `require` system for all imports in the main module,
 - turn off the new feature for renderers called `contextIsolation`,
 - remove the enabling of `enableRemoteModule` because it no longer exists,
 - remove `electron-util` from all renderers (it's no longer compatible with renderer processes).  This change effected:
   - checks for the OS, and
   - checks for whether we're in development mode.
 - `shell.moveItemToTrash` became `shell.trashItem`, and
 - the way that the menu is built needed to re-implemented with an injected function rather than a call to `require`.
 
 # Other changes
 - Fixed the point that we ask the user to log in when the session expires.


src/isOS.js
```diff
@@ -0,0 +1,13 @@
+export const isWindows = () => {
```
Quiescent: Reading through this PR, I'm not sure that these checks are good enough.
Back to the drawing board for these.

PR #761, opened 2022-03-04
Import scrivener

jeanawhou:
@cameronsutter @Quiescent this still has a lot of todos 😅 but for now its working and can upload simple scrivener project. 
I will work on a project that has two books, so folder structure im looking is each manuscript folder will be 1 book (let me know if that doesn't sound right 🙇 )
I will attach the demo on the slack because the my video is too large (i apologize for my accent, voice and broken english xD)


lib/plottr_components/src/components/dashboard/files/FilesHome.js
```diff
@@ -37,6 +45,21 @@ const FilesHomeConnector = (connector) => {
     }
   }
 
+  function scrivenerImportDialog() {
+    const title = t('Choose your Scrivener project')
+    const filters = [{ name: t('Scrivener file'), extensions: ['scriv'] }]
+    const properties = ['openFile', 'createDirectory']
```
Quiescent: These options don't work for me.  I need `'openDirectory'` in the list for it to work on Linux, but when I add it, I have to remove the `extensions` filter on the above line.
@cameronsutter can you comment here?
What have you used to open folders in the past (Scrivener files are a bit of an edge case).

lib/plottr_components/src/components/dashboard/files/FilesHome.js
```diff
@@ -37,6 +45,21 @@ const FilesHomeConnector = (connector) => {
...
+    const files = showOpenDialogSync({ title, filters, properties })
+    if (files && files.length) {
+      const importedPath = files[0]
+      if (importedPath.toLowerCase().includes('.scriv')) {
+        return importedPath
+      } else {
+        return null
```
Quiescent: Perhaps we should tell the user that they made a mistake?
You could add another case to the `error` reducer.

lib/plottr_components/src/components/dashboard/files/NewFiles.js
```diff
@@ -78,15 +78,20 @@ const NewFilesConnector = (connector) => {
...
                   active: activeView == 'import',
                 })}
-                onClick={doImport}
               >
-                <FaRegSnowflake />
-                <div>{t('Import Snowflake Pro')}</div>
-              </div>
+                <Dropdown.Toggle noCaret>
+                  <FaRegSnowflake />
+                  <div>{t('PRO Import')}</div>
```
Quiescent: I think this should just be 'Import' because it'll be on non-pro too.
@cameronsutter, can you please comment here?

lib/plottr_components/src/connections/pltr.js
```diff
@@ -129,6 +129,7 @@ const pltrTypeSpecs = {
     readFileSync: PropTypes.func.isRequired,
     moveItemToTrash: PropTypes.func.isRequired,
     createFromSnowflake: PropTypes.func.isRequired,
+    // createFromScivener: PropTypes.func.isRequired,
```
Quiescent, line 132: Please uncomment this line.
It's an important check to avoid run time errors in prod.

main/modules/importer/scrivener/importer.js
```diff
@@ -0,0 +1,764 @@
+import fs from 'fs'
+import xml from 'xml-js'
+import { t } from 'plottr_locales'
+import log from 'electron-log'
```
Quiescent: Please delete this line.

src/files.js
```diff
@@ -54,6 +71,36 @@ export const newFile = (
   })
 }
 
+export const newFileFromScrivener = (
```
Quiescent: I'm not sure that this is used anymore.  Could you please check?

src/files.js
```diff
@@ -31,6 +31,23 @@ export const newEmptyFile = (fileName, appVersion, currentFile) => {
   }
 }
 
+export const importScrivenerState = (fileName, appVersion, currentFile) => {
```
Quiescent: I'm also not sure that this is used.

src/common/exporter/scrivener/importers/notes.js
```diff
@@ -0,0 +1,14 @@
+export const createNotesFromScriv = (defaultNote, state) => {
+  const notes = state.sections.find((i) => {
+    return i.title === 'Notes'
+  })
+  return notes.children.map((item, key) => {
+    return (defaultNote = {
```
Quiescent: I'm not sure why you're doing this assignment(?)

package.json
```diff
@@ -51,12 +51,15 @@
     "electron-updater": "4.3.9",
     "electron-util": "^0.17.2",
     "electron-window-state": "^5.0.3",
+    "fs.promises": "^0.1.2",
```
Quiescent: This should be built in.  What version of node are you running locally?

main/modules/importer/scrivener/importer.js
```diff
@@ -0,0 +1,764 @@
...
+      }
+    })
+    return { lineId: rtfContents[0] && rtfContents[0].lineId ? rtfContents[0].lineId : 1 }
+  } else if (isSection && rtf.contents && rtf.contents.length) {
+    return rtf.contents
+  }
+}
+
+function generateNotes(currentState, json, bookId, files, isNewFile) {
+  json.map((item) => {
```
Quiescent: If you `map` an array for side-effect (i.e. you throw away the result of mapping), then please use `forEach` instead.
(`map` allocates a new array and also implies to the reader that you're interested in the result.)

main/modules/importer/scrivener/importer.js
```diff
@@ -0,0 +1,764 @@
...
+          }),
+        }
+      }, withChildStoryLinesAdded)
+    }, withNewBeats)
+
+  return withNewCardsAndNewBeats
+}
+// Object, Array <any>, Object, Int -> Object { txtContent: {}, rtfContent: {} }
+function mapMatchedFiles(newJson, matchFiles, fileContents, bookId, isSection) {
+  matchFiles.flatMap((file) => {
```
Quiescent: Similar problem here.  We discard the value, but we're using `flatMap`.
`flatMap` should be reserved for instances where we:
 - are interested in the result of `flatMap`ping, and
 - the function that we pass to `flatMap` produces an array itself.  (i.e. if we used ordinary `map` it would produce the type `[[Any]]`, but instead it'll create a "flattened" array `[Any]`.
 
 I think we should use `forEach` here again.

main/modules/importer/scrivener/importer.js
```diff
@@ -0,0 +1,764 @@
...
+    const noRTF = !file.filePath.includes(file.filePath.endsWith('.rtf'))
```
Quiescent: I don't understand what this line does.  I read it as "does the file's filePath include a boolean which is true if the filePath ends with '.rtf'?!

jeanawhou: Thank you ed! PR updated!

Quiescent (approved):
Looks good.  LGTM!

lib/plottr_components/src/components/dashboard/files/FilesHome.js
cameronsutter: That's a good point. Do we make them choose the .scriv or can they choose the project folder? On Mac, the project folder is displayed as a file, not a folder because it's a "package."

Maybe this should be Mac/Win/Linux dependent

lib/plottr_components/src/components/dashboard/files/NewFiles.js
cameronsutter: Yup! It should say "Import"

We'll also need to change the icon, but don't worry about that. I'll get Ryan to make that choice

main/modules/importer/scrivener/from_html.js
```diff
@@ -0,0 +1,75 @@
+import jsdom from 'jsdom'
```
cameronsutter, line 1: is this file different from the other "from_html.js" that we have? It looks the same

Unless it's slightly customized for scrivener, I don't think we need to copy this into the "importer/scrivener" folder. We can just use the other one

PR #765, opened 2022-03-09
Ask for project name

Quiescent:
# Motivation

When users create new projects, there's a fair chance that they already know what they want to call it.
In this PR, we allow the user to give the new file a name when it's created.


jeanawhou (approved):

Quiescent:
Don't merge yet: I just realised that we need to check whether we're overwriting another file in the temp directory when we create the file.

Quiescent:
> Don't merge yet: I just realised that we need to check whether we're overwriting another file in the temp directory when we create the file.

I've since fixed this 👍

PR #768, opened 2022-03-10
Firebase worker

Quiescent:
# Motivation

To support Live RCE, we adopted a CRDT.
Merging and various other operations on the CRDT are expensive, so we elected to offload those to a thread.
In this PR, we start the migration to a CRDT by moving all application-level network traffic to a thread (i.e. the whole of our Firebase API).

# ⚠️ Login is now done via a popup ⚠️
 
I found during development that the Firebase queue would lock up when Plottr initialised it on both the worker thread and the main thread.
So part of this PR involves moving the login component to a popup window to run on the main thread on its process without locking the queue for the app window's process.


src/_firebase-worker.js
```diff
@@ -0,0 +1,465 @@
...
+        })
+      })
+      unsubscribeFunctions.set(messageId, unsubscribe)
+      return
+    }
+    case LISTEN_TO_CUSTOM_TEMPLATES: {
+      const { userId } = messagePayload
+      const unsubscribe = listenToCustomTemplates(userId, (result) => {
+        self.postMessage({
+          type: typeToReplyType(LISTEN_FOR_RCE_LOCK),
```
jeanawhou: @Quiescent im not sure if this is typo error or intentional 😅

Quiescent: Yup.  This definitely looks like a mistake XD

PR #769, opened 2022-03-14
Bugfix/import scrivener

jeanawhou:
@cameronsutter @Quiescent 

on this PR i added the following:
- fix for empty/unarranged custom attributes
- added tags
- added categories

The video demo i will post on the slack #dev-desktop channel 😅 
Thank you 🙇


lib/pltr/v2/slate_serializers/to_html.js
```diff
@@ -18,6 +18,12 @@ export const rtfToHTML = (arrayBuffer) => {
         for (let j = 0; j < nodeList.length; ++j) {
           result.push(nodeList[j])
         }
+
+        if (!nodeList.length) {
```
Quiescent, line 22: @jeanawhou, could this create a whole bunch of empty paragraphs?
What if we moved it outside the outer loop and checked whether `result` is empty?
If the `result` is empty, we can add a node to the result list exactly once.

main/modules/importer/scrivener/importer.js
```diff
@@ -628,27 +660,55 @@ function generateNotes(currentState, json, bookId, files, isNewFile) {
...
-        (obj, item) => Object.assign(obj, { [Object.keys(item)[0]]: Object.values(item)[0] }),
-        {}
+      const content = generateCustomAttribute(
+        { contents: fileContents.rtfContents },
+        currentState,
+        name
       )
+      const customAttributes = content
+        // remove three dots placeholder if key
+        .filter((item) => Object.keys(item)[0] != '...')
```
Quiescent: Doing something like `Object.keys(x)[0]` is only valid if there's exactly one key on the object.
Otherwise, you're actually choosing a key potentially at random(!)
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#deleted_added_or_modified_properties

Quiescent: Also see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values
(It says that `Object.values` -and it's the same for `Object.entries` and `Object.keys`- iterates over the object's properties in the same order as `for...in`.)

Quiescent: So, I suppose that my question is: does `item` have only one key?

main/modules/importer/scrivener/importer.js
```diff
@@ -661,21 +721,49 @@ function generatePlaces(currentState, json, bookId, files, isNewFile) {
         }
       }
       createNewPlace(currentState.places, values)
+      createCustomAttributes(currentState, content, 'places')
+    } else if (children['BinderItem']) {
+      // for the manuscript we assign the parent folder of the card as the line
+      // we could assign the parent folder as the category
+      // WDYT @ed @cameron ?
```
Quiescent, line 735: I'm not sure that I follow.
Aren't we creating a place?
How does a place have a line?

main/modules/importer/scrivener/importer.js
```diff
@@ -713,24 +840,39 @@ function generateCustomAttribute(fileContents, name) {
...
       .map((attributes) => {
         const newAttr = Object.keys(attributes)[0]
-        const attributeExists = currentState.customAttributes.characters.find(
-          (attr) => attr.name == newAttr
+        const attributeExists = currentState.customAttributes[section].find(
+          (attr) => attr.name.toLowerCase().trim() == newAttr.toLowerCase().trim()
         )
 
-        if (!attributeExists || !currentState.customAttributes.characters.length) {
+        const tagAttribute = newAttr.toLowerCase().trim() == 'tags'
```
Quiescent: It looks like there are a bunch of unused variables here.
Is that intentional?

main/modules/importer/scrivener/importer.js
jeanawhou: Yes @Quiescent because it is an array of objects

main/modules/importer/scrivener/importer.js
jeanawhou: No ed, I forgot about this one 😅

main/modules/importer/scrivener/importer.js
Quiescent: Then we might instead use `item[0]`, right?

main/modules/importer/scrivener/importer.js
jeanawhou: It failed when i uploaded a scrivener file that has _Places with a subfolder_ because i think i created the folder on the `Places` wrong that's why it created another subfolder instead of like a child container (that looks like have an index card see the image attach on the places folder the `another place` folder) and it didn't match its uuid to its child(the one that has index card). 

Inside the folder with index card, it matched the child and its content. 
That's why i added this check
image

main/modules/importer/scrivener/importer.js
jeanawhou: I get undefined 😅 inside the filter and map function when i use item[0]

Quiescent: Then `item` must not be an array. :)
What does `item` actually look like?

jeanawhou: I mean the content its an array of objects, the item is an object 😅

main/modules/importer/scrivener/importer.js
Quiescent: Ah!  I think I understand.
So you're asking whether it makes sense to use that subfolder as a category.
I'd say: yes, *BUT*  I don't know Scrivener very well.
Is this something that people usually do in Scrivener?
(@cameronsutter, do you have any idea if this is common in Scrivener files?)

main/modules/importer/scrivener/importer.js
Quiescent: Right.  That's fine :+1: 
The problem here is the assumption that `Object.keys(item)[0]` will produce the correct key.
It might do that in your tests, but the standard doesn't guarantee this behaviour.
Does `item` have a known schema?

jeanawhou: I forgot to add a signature here https://github.com/cameronsutter/plottr_electron/blob/1ddecff8cc6b7db99178feccb72c7467439325bb/main/modules/importer/scrivener/importer.js#L821

This function should have a signature of 
`// Object, Object, String -> [Object]`
It will return an array of objects

lib/pltr/v2/slate_serializers/to_html.js
jeanawhou: @Quiescent this only will create a new span if the nodeList is empty, not the entire result (1st image). I added this check so that if the user didn't put any description for the custom attribute but has title or it is blank it will have a placeholder of 3 dots and it will not break custom attribute and get the next text (which is Role see 2nd image) 😅 

Each nodeList will be a custom attribute and then its description consecutively 
So the nodeLists array there on the local values of the first image is 

first nodeList [0] = **title**
first nodeList [1] = **first custom attribute**
next nodeList **first custom attribute description**

next nodeList **2nd custom attribute**
next nodeList **2nd custom attribute description**, and so on...
What do you think ed?

Screen Shot 2022-03-15 at 4 54 27 PM

image

jeanawhou: Sorry this is the right nodeLists for that three little pigs i was uploading 
image

Quiescent: Right!  Thanks for explaining.
I think that's fine.  Let's see what Troy and others think about how it looks in practice.

main/modules/importer/scrivener/importer.js
Quiescent: @jeanawhou thanks, but we might be missing each other here.
I'm trying to resolve the problem caused by using `Object.keys(item)[0]` <--
Here, the intention seems to be to read the first key of an object.
What I'm saying is that `Object.keys(item)[0]` could give you *any* key(!)
So let's figure out what *`item`* looks like (i.e. what are its keys) so that we can avoid using `Object.keys` (because using `Object.keys` is probably a bug).

jeanawhou: This how an item looks (1st image)
image

And it will depend on the custom attribute they wrote

on the scrivener
It will look llike this 
image

On the rtf it will look like this 
image

Quiescent: Thanks for explaining :+1:  :slightly_smiling_face:
Ok.  So items are **by definition** single attribute objects(?)
I'm happy to keep it as is as long as we add a comment at the top of the function that makes the assumption explicit.
i.e. // ASSUMPTION: `item`s are single-key objects with a name that corresponds to an entity's property and a value which corresponds to that property's value.

jeanawhou: I see thank you ed

Quiescent (approved):
LGTM and get a beta build out for testing!

PR #770, opened 2022-03-17
Cards toast notification

jeanawhou:
Ticket: https://airtable.com/appfbr7vlSqIfZ9Sn/tblAmQkxs32m46J4N/viwdiMn4iZIDn3TH1/recyQWhf9likE5Laq?blocks=hid

Demo

https://user-images.githubusercontent.com/19387007/158794052-6e206160-b08c-496a-b924-abd273777e0c.mp4


Quiescent:
From the demo, this is looking great! 🙂 

I didn't expect the notification when a scene card moves on the same timeline.
(I think the spec might not have been specific about that, though.)

@cameronsutter, what do you think about the notification when a card is moved on the same timeline.
(I think it might be slightly too noisy.)

lib/plottr_components/src/components/timeline/TimelineTable.js
```diff
@@ -33,6 +34,12 @@ const TimelineTableConnector = (connector) => {
     state = {
       tableLength: 0,
       hovering: null,
+      toastProps: {
```
Quiescent: I think that this should be in `redux` rather than state on the TimelineTable.

lib/plottr_components/src/components/timeline/TimelineTable.js
```diff
@@ -57,6 +64,12 @@ const TimelineTableConnector = (connector) => {
 
     componentDidUpdate() {
       this.setLength()
+
+      if (this.state.toastProps.visible) {
```
Quiescent: This is the wrong place for this timeout to be registered -- what happens when there are more updates before the timer expires?
If the state is in `redux`, this should all become clearer.

lib/plottr_components/src/components/timeline/TimelineTable.js
```diff
@@ -273,6 +286,60 @@ const TimelineTableConnector = (connector) => {
       }
     }
 
+    handleCloseToast = () => {
```
Quiescent, line 285: This should be an action.

lib/plottr_components/src/components/timeline/TimelineTable.js
```diff
@@ -273,6 +286,60 @@ const TimelineTableConnector = (connector) => {
...
+      this.setState({
+        toastProps: { visible: false, action: null, cardId: null, bookId: null },
+      })
+    }
+
+    handleShowToast = (action, cardId, bookId) => {
```
Quiescent: This should also be an action.

lib/plottr_components/src/components/timeline/TimelineTable.js
```diff
@@ -273,6 +286,60 @@ const TimelineTableConnector = (connector) => {
...
+      this.setState({ toastProps: { visible: true, action, cardId, bookId } })
+    }
+
+    getToastMessage = () => {
```
Quiescent: This should be a selector.

lib/pltr/v2/store/initialState.js
```diff
@@ -70,6 +70,11 @@ export const ui = {
   timeline: {
     size: 'large',
   },
+  toastProps: {
```
Quiescent: Your choice of where to put this state is logically sound; however, the `ui` key gets saved to files.
Could you please put it on a different reducer?
Our system reducers (state that isn't saved to file) are:
 - 'error',
 - 'permission',
 - 'project',
 - 'actions',
 - 'client',
 - 'license',
 - 'knownFiles',
 - 'templates',
 - 'settings',
 - 'backups',
 - 'applicationState', and
 - 'imageCache'.
(As copy-pasted from `systemReducers.js`).

If none of these are a good place, then please feel free to add another reducer (`notifications` perhaps?).

Quiescent:
Well done on this change.  It looks rock solid. :+1: 
I left a comment about an unintended consequence of putting the state on `ui`.
Once you resolve this, we'll be good to go! :rocket: :moon:

jeanawhou:

lib/plottr_components/src/components/timeline/CardDialog.js
```diff
@@ -602,6 +603,7 @@ const CardDialogConnector = (connector) => {
     getTemplateById: PropTypes.func.isRequired,
     customAttributes: PropTypes.array.isRequired,
     uiActions: PropTypes.object.isRequired,
+    notificationActions: PropTypes.func.isRequired,
```
Quiescent: I think this might be an object(?)

lib/plottr_components/src/components/timeline/TimelineTable.js
```diff
@@ -449,7 +449,8 @@ const TimelineTableConnector = (connector) => {
     cardActions: PropTypes.object,
     beatActions: PropTypes.object,
     books: PropTypes.object.isRequired,
-    toastProps: PropTypes.object,
+    toast: PropTypes.object,
+    notificationActions: PropTypes.func,
```
Quiescent: I think this is also an object.

Quiescent:
Final todo: every system reducer must have its key name and actions types added to ‘systemReducers.js’ to make sure that they don’t cause synchs to Firebase, can’t be undone and don’t get saved to Plottr files.

Quiescent (approved):
Good work.  LGTM!  You’re on a roll! :)

PR #774, opened 2022-03-30
Fix exporting to Scrivener when we need to delete an existing file

Quiescent:


src/app/_index.js
```diff
@@ -2,6 +2,7 @@ import path from 'path'
 import { store } from 'store'
 import { ipcRenderer } from 'electron'
 import { dialog, getCurrentWindow } from '@electron/remote'
+import { rm } from 'fs'
```
cameronsutter: do we need this anymore?

src/app/_index.js
```diff
@@ -101,6 +102,7 @@ ipcRenderer.on('export-file-from-menu', (event, { type }) => {
     logger,
     exportSaveDialog,
     MPQ,
+    rm,
```
cameronsutter: wouldn't we want to use the rmRF here instead?

cameronsutter (commented):
Just a couple questions. I'm probably just not following the code well enough

src/app/_index.js
Quiescent: You're right.  Good catch.  It should fail when we export via the file menu at the moment.
I'll double-check that that's the case for sanity's sake.

src/app/_index.js
Quiescent: Yup

PR #785, opened 2022-04-18
Bugfix/import scrivener

jeanawhou:
@Quiescent @cameronsutter I added the ability to accept characters, places and tags from cards/notes.rtf and also fix the contenating values for the RCE.

I will add the demo on the #dev-desktop channel 😅


lib/plottr_import_export/src/importer/scrivener/importer.js
```diff
@@ -645,17 +757,108 @@ function mapMatchedFiles(newJson, matchFiles, fileContents, bookId, isSection) {
...
+          isSection
+        )
         Object.assign(fileContents.rtfContents, newRTFContent)
       }
     }
   })
   return fileContents
 }
 
+function getManuscriptSectionItems(currentJson, RTFContents, bookId) {
```
Quiescent: This function needs tests.
It's riddled with assumptions and heuristics, and we must verify that it works as we intended it to.

lib/plottr_import_export/src/importer/scrivener/importer.js
```diff
@@ -645,17 +757,108 @@ function mapMatchedFiles(newJson, matchFiles, fileContents, bookId, isSection) {
...
+          [strippedSectionName]: sectionItems.includes(strippedSectionName)
+            ? sectionIds
+            : sectionText,
+        }
+      } else return null
+    } else return null
+  })
+}
+
+function filterRTFContents(RTFContents) {
```
Quiescent: This function needs tests and a better name.
The name should describe the intention, not the function we called on `RTFContents`.
How about `keepNonDuplicatedNonEmptyText` ?
Also: what if a user repeats a line verbatim?

Quiescent (changes requested):
Good work.

Before we go ahead and merge this: the functionality in `importer.js` is very complicated, and you are now *the* person with a complete understanding of it.
Given:
 a) how important this feature is, and
 b) how complicated the logic is,
I'm going to insist that you add unit tests for every function in this file.

jeanawhou:
@Quiescent @cameronsutter i added a unit test for all the functions in scrivener importer except for the functions with file handling. I've copied the jsonified example3.scriv and that's what i used to create the tests. Should i also add unit test for file handling?

Quiescent:
> @Quiescent @cameronsutter i added a unit test for all the functions in scrivener importer except for the functions with file handling. I've copied the jsonified example3.scriv and that's what i used to create the tests. Should i also add unit test for file handling?

I'm taking a look.  In principle, I'm fine with not unit testing the file handling. :+1:

lib/plottr_import_export/src/importer/scrivener/__tests__/importer.test.js
```diff
@@ -0,0 +1,911 @@
...
+    it('should return a json', () => {
+      const storyLines = withStoryLineIfItExists(
+        manuscript,
+        currentState,
+        BOOK_TITLE,
+        bookId,
+        isNewFile,
+        relevantFiles
+      )
+      expect(isAnObject(storyLines)).toBeTruthy()
```
Quiescent, line 358: Is there any assertion we can make about the members of the object?
i.e. does it look like RCE content?
If it does, then it would be useful to verify that it has the right members as well as being an object.

lib/plottr_import_export/src/importer/scrivener/__tests__/importer.test.js
```diff
@@ -0,0 +1,911 @@
...
+      expect(similarCards[0].id).not.toEqual(similarCards[1].id)
+      expect(currentCards).toHaveLength(4)
+    })
+  })
+
+  describe('createNewLine', () => {
+    const newLine = createNewLine(currentState.lines, { title: TEST_STRING_1, id: 1 }, bookId)
+
+    it('should return the line id', () => {
+      expect(typeof newLine).toBe('number')
```
Quiescent: Does the line id change between test runs?
If not, perhaps we should check that it's a specific number.

Quiescent:
This PR was a massive undertaking.  Thank you for putting in those tests! :raised_hands:
Please see the two minor comments I had.

jeanawhou:

Quiescent (approved):

Quiescent (approved):
LGTM!

PR #790, opened 2022-04-28
Socket server process

Quiescent:
# Motivation

To create a genuinely asynchronous server to which we can offload heavy lifting.

# Approach

If Electron creates a window using `fork`, that window starts with an environment variable `NODE_PROCESS` (or something akin to that) set to true.
These processes are asynchronous, and Electron can't block them when it's doing its work.

In this PR, we create a node process on startup that listens for web socket connections.
Its structure is very similar to the firebase worker, but it communicates over a web socket instead of the web worker protocol.
When this PR is merged, the worker will support the `ping` and `rm_rf` messages.

See `connected-components.js` for an example usage (search for `whenClientIsReady`.)


jeanawhou (approved):
LGTM :)

main/server/init.js
```diff
@@ -0,0 +1,42 @@
...
+    if (attempts >= MAX_ATTEMPTS) {
+      return Promise.reject(`Failed to bind socket server after ${MAX_ATTEMPTS} attempts.`)
+    }
+
+    return new Promise((resolve, reject) => {
+      const randomPort = START_PORT + Math.floor(1000 * Math.random())
+      log.info(`Starting socket server on port: ${randomPort}`)
+
+      const server = fork(join(__dirname, './socketServer.bundle.js'), [randomPort])
+      server.on('close', (code) => {
```
cameronsutter, line 21: if this happens sometime after the initial startup, it looks like it restarts the server on a new port. Does the rest of the app know that it's now on a new port?

main/server/init.js
```diff
@@ -0,0 +1,42 @@
...
+        log.warn(`Socket server died with code: ${code}`)
+        if (code === 1) {
+          log.warn(`Restarting the server on a new port.`)
+          attempts++
+          return attemptAStart()
+        }
+        log.error(`Failed with an unhandled error.  Killing the server.`)
+        return reject(new Error(`Socket worker died with unhandled error code: ${code}`))
+      })
+      server.on('message', (message) => {
```
cameronsutter, line 31: Am I reading this correctly that the server isn't "up and running" and considered ready until it receives its first message from the socket worker (I'm assuming that's in a project worker)? Is that going to cause any problems with the timing of different pieces at startup and was that intentional?

cameronsutter: And I think i missed where in the code it sends the 'ready' message after startup

cameronsutter (commented):
Just a question

main/server/init.js
Quiescent: No, good point.  The plan was to broadcast the change, but I never got around to it.
I'll create a follow-up PR with this and any other changes.

main/server/init.js
Quiescent: So yes, it considers this as part of the critical path of booting up the app.
We don't need a project open, though.  Here, the main process waits for the forked process to tell it that it started the socket server.

See: https://github.com/cameronsutter/plottr_electron/blob/staging/main/server/server.js#L58
That line uses interprocess communication rather than a socket to let the parent process know that the socket server is up and running on the child process.

PR #799, opened 2022-05-20
fix verification choice view for old version of small screens

jeanawhou:
@Quiescent @cameronsutter this is the fix for the responsivity for the verify view on a mac with intel chip


Quiescent (approved):
Looks good!  LGTM!

PR #801, opened 2022-05-26
Extract react bootstrap

Quiescent:
# Motivation
Our bootstrap version is ancient (or at least it is in tech terms.)
It's caused us a lot of headaches in bugs, made certain features harder to implement than they should be, and tied us to old dependencies.
We can't upgrade `core-js` or React and many other dependencies unless we solve this problem.

# Solution
Copy all of the components that we use from the version of `React-Bootstrap` that we're using in-tree.
Remove dependencies and other problems in the newly-copied components.
Gradually delete the CSS etc., that we don't use right now.

# Alternative
We could have tried to upgrade `React-Bootstrap` itself.
We tried this route initially.
We realised that the styles and APIs changed without any material benefit to Plottr.
i.e. we'd spend a lot of time undoing the changes that the upgrade made, just so that we didn't transitively depend on packages that prevent upgrading.

# Testing
I suggest a full regression test for this change.
Anything from a Dropdown to the colour of a button could have changed inadvertently.
I've done my best to interact with all of the elements in the project, but I might've missed something.


Quiescent:
I'm working my way through the linting errors.
Please don't merge this until I give the go-ahead on that front.

Quiescent:
I fixed all but two linting errors.
Both of which are `findDOMNode` usages.
I think that the `DropDown` won't close right now (as of refactoring* it to a function component).

* Well, that was the intention anyway :stuck_out_tongue_closed_eyes:

cameronsutter (approved):

jeanawhou (approved):

Quiescent:
I've fixed almost all of the warnings.
There are a couple of harmless ones floating about, and the signal to noise ratio is greatly improved.

PR #803, opened 2022-05-31
Move saving files to socket server

Quiescent:
# Motivation
If the main process is blocked, the renders are also blocked.
We've observed instances where the entire app freezes in production, and we're not sure why.
Our understanding is that something is blocking the main process.
We've devised a solution: run potentially blocking and long-running tasks on another server that *cannot block Electron*.

# This PR: Move Save to The Socket Server
We move the function `saveFile` from the main process to the socket server in this PR.

Secondarily, the PR also improves how the architecture of the app.
Namely, the PR moves the client to a shared folder (shared between renderers and the main process) and adds a context provider for communicating with the main process.
I envisage that all side effects relating to the file system may be provided by the new context so that we can change side-effecting functions more easily.


src/mainIntegrationContext.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+const MainIntegrationContext = React.createContext({
+  saveFile: (filePath, data) => {
```
jeanawhou, line 4: @Quiescent im not sure what this means. I haven't tried creating React Context 🥲

jeanawhou (approved):
LGTM :)

src/mainIntegrationContext.js
Quiescent: It's a way to communicate values through components without explicitly passing them through the props of each one.
(In that way, it's a little bit like redux).
Here are the docs: https://reactjs.org/docs/context.html

e.g.
```
const ExampleContext = React.createContext({ someValueWeWantToPassDown })
```

A component must provide the context, and another must consume it.

Here's a consumer:

```
const SomeInnerComponent = ({}) => {
  return (
    <ExampleContext.consumer>
      {({ someValueWeWantToPassDown }) => {
          return (
            <span>Message passod down without props(!): {someValueWeWantToPassDown}</span>
          )
      }}
    </ExampleContext.consumer>
}
```

And here, we provide the context so that anything inside the tree that the provider surrounds can access the context.

```
const AComponentForDemoPurposes = () => {
  return (
    <ExampleContext.provider value={{someValueWeWantToPassDown: 'hooray!'}}>
      <div>
        <Form>
          ...
        </Form>
        <SomeComponent>
          <SomeInnerComponent />
        </SomeComponent>
      </div>
    </ExampleContext.provider>
}
```

jeanawhou: Thank you @Quiescent for the explaination 🙇

src/mainIntegrationContext.js
```diff
@@ -0,0 +1,9 @@
...
+    throw new Error(`Couldn't save file at ${filePath} with data: ${data} because `)
+  },
```
jeanawhou, line 6: @Quiescent here, its only throwing an error when the saveFile is called. I didn't get it 😅

cameronsutter (commented):
Looks good to me.

It looks like `saveOfflineFile` was left for a later time, is that right?

Do you think this one will block the renderer processes and is it feasible to implement without all the hairy changes required for some of the other things we talked about?

src/mainIntegrationContext.js
Quiescent: This is the default behaviour.  It'll only happen when the programmer forgets to give a value to `saveFile` in the context provider.
i.e. if someone forgets to define it in the provider, then that's a mistake and the error should be very visible during development so that we don't accidentally take it into production.
See: https://github.com/cameronsutter/plottr_electron/pull/803/files#diff-bf2d025de96b30a546c2e33beafd0daf68e8b64f2fcc74b6c6c8a135ff6c37abR23

Quiescent:
> Looks good to me.
> 
> It looks like `saveOfflineFile` was left for a later time, is that right?
> 
> Do you think this one will block the renderer processes and is it feasible to implement without all the hairy changes required for some of the other things we talked about?

`saveOfflineFile` looks safe.  It would've blocked the main process.  I'll take a look at it now.
`saveBackup` is a bit hairier because it depends on electron a whole lot.

PR #805, opened 2022-06-02
Bugfix/import scrivener

jeanawhou:
@Quiescent @cameronsutter i was able to successfully import the test files from the dev-desktop channel on slack.

### Fractured test import 

- The chapter titles are just too long
- characters are not organized. Adding description or notes needs a label (i.e. description - 'im a description of a character') as it treats everything from the rtf as attributes.

### A Coroners Tale 
- overall ok

### The turn
- same problem with fractured test import for the section items. The place needs a label if its a description, note or something else so it will know where place the values.

I've attached a zip files with the .pltr after the test files where imported. Demo will be posted on dev-desktop
[test scrivener files.zip](https://github.com/cameronsutter/plottr_electron/files/8820068/test.scrivener.files.zip)


main/modules/files.js
```diff
@@ -434,19 +438,29 @@ function createFromScrivener(importedPath, sender, isLoggedIntoPro) {
...
-    importedJsonPromise.then((importedJson) => {
-      sender.send('create-plottr-cloud-file', importedJson, storyName)
-    })
+    importedJsonPromise
+      .then((importedJson) => {
+        sender.send('create-plottr-cloud-file', importedJson, storyName)
+      })
+      .catch((error) => {
+        log.error(error)
+        sender.send('finish-creating-local-scrivener-imported-file')
```
Quiescent: This line doesn't seem right: we ran into an error and simply stopped without letting the user know what happened?
We should probably add a second listener in `_index.js` to handle errors.

main/modules/files.js
```diff
@@ -420,6 +420,10 @@ function createRTFConversionFunction(sender) {
 }
 
 function createFromScrivener(importedPath, sender, isLoggedIntoPro) {
+  if (importedPath.error) {
+    log.error(importedPath.error)
+    return sender.send('finish-creating-local-scrivener-imported-file')
```
Quiescent: See my previous comment.  We need to let the user know that something went wrong.

src/app/_index.js
```diff
@@ -285,15 +285,20 @@ ipcRenderer.on('create-plottr-cloud-file', (event, json, fileName) => {
...
+      openFile(`plottr://${fileId}`, fileId, false)
+      // Fixme: this could have been called to create from a snowflake
+      // file too(!)
+      store.dispatch(actions.applicationState.finishScrivenerImporter())
+      closeDashboard()
+      return fileId
+    })
+    .catch((error) => {
+      logger.error(error)
+      store.dispatch(actions.applicationState.finishScrivenerImporter())
```
Quiescent: Here's another case that we should let the user know that something failed.

main/modules/files.js
```diff
@@ -434,19 +438,29 @@ function createFromScrivener(importedPath, sender, isLoggedIntoPro) {
...
+  return importedJsonPromise
+    .then((importedJson) => {
+      return saveToTempFile(importedJson, storyName).then((filePath) => {
+        const fileId = addToKnownFiles(filePath)
+        openKnownFile(filePath, fileId)
+        sender.send('finish-creating-local-scrivener-imported-file')
+      })
+    })
+    .catch((error) => {
+      log.error(error)
```
Quiescent: And here again with letting the user know.

Quiescent:
I'd like to see all three failing files checked into the repo.
If you could write a test that checks that the whole process works for each file:
 - imports a Scrivener file, and
 - checks it against a version that you create today,
that would make me much more confident about changes going forward.

From now on, when we get a bug like this, please could you do the following:
 - create a branch and commit the failing files,
 - write a test that tries to import the file (the test should fail otherwise we've failed to replicate the bug),
 - commit those changes and push them to the branch on origin,
 - fix the problem and turn the test from red to green.

jeanawhou:
> I'd like to see all three failing files checked into the repo. If you could write a test that checks that the whole process works for each file:
> 
> * imports a Scrivener file, and
> * checks it against a version that you create today,
>   that would make me much more confident about changes going forward.
> 
> From now on, when we get a bug like this, please could you do the following:
> 
> * create a branch and commit the failing files,
> * write a test that tries to import the file (the test should fail otherwise we've failed to replicate the bug),
> * commit those changes and push them to the branch on origin,
> * fix the problem and turn the test from red to green.

@Quiescent i was able to import the 3 files successfully, they just didn't have the content from the main body (see highlighted part of the screenshot)

Screen Shot 2022-06-03 at 4 07 07 PM


Should i jsonified them 3? Test them commit that if they passed or failed (on a different branch)?

Quiescent:
> > I'd like to see all three failing files checked into the repo. If you could write a test that checks that the whole process works for each file:
> > 
> > * imports a Scrivener file, and
> > * checks it against a version that you create today,
> >   that would make me much more confident about changes going forward.
> > 
> > From now on, when we get a bug like this, please could you do the following:
> > 
> > * create a branch and commit the failing files,
> > * write a test that tries to import the file (the test should fail otherwise we've failed to replicate the bug),
> > * commit those changes and push them to the branch on origin,
> > * fix the problem and turn the test from red to green.
> 
> @Quiescent i was able to import the 3 files successfully, they just didn't have the content from the main body (see highlighted part of the screenshot)
> 
> Screen Shot 2022-06-03 at 4 07 07 PM
> 
> Should i jsonified them 3? Test them commit that if they passed or failed (on a different branch)?

@jeanawhou I know that it's working now :)
I want to future proof this change and make it less likely to break again.
We can do two things to ensure that:
1. Add tests for these files so that we don't run into this exact problem again.
2. Add tests every time that we run into a problem like this (using the process of implementing failing tests, then committing them while they're failing(!), then making them pass and committing again) so that we're more confident in the fix and prevent the same problem from arising in the future.

You've already fixed this, so I'm only asking for tests to prevent this from breaking again.
i.e. for this PR, I'd like each *entire Scrivener file/folder* to be checked in so that we can run tests that go through the whole process.
i.e. the test should behave precisely the same way as production does.
*Then* check in the JSONified result of importing the scrivener file *as well* and test that the result of importing the original Scrivener file is identical to the JSONified result.

jeanawhou:
@Quiescent the import didn't fail.

But there's couple of test that did failed because the test is not expecting an empty character items :sweat_smile: . After i added checks for the nulls the unit test passed, all three scrivener files (the jsonified) pass the unit test. 

I'm not sure which test scrivener should i choose :sweat_smile: 
Or should i update the unit test and put the original file plus the 3 file to an array and iterate them to all the test? :sweat_smile:

Quiescent:
> @Quiescent the import didn't fail.
> 
> But there's couple of test that did failed because the test is not expecting an empty character items sweat_smile . After i added checks for the nulls the unit test passed, all three scrivener files (the jsonified) pass the unit test.
> 
> I'm not sure which test scrivener should i choose sweat_smile Or should i update the unit test and put the original file plus the 3 file to an array and iterate them to all the test? sweat_smile

@jeanawhou, I'm glad that the tests did pick it up.  (Good job! :) )

The type of test that I'm asking for isn't a unit test.
It would go in a separate file.
Its primary use is to check the plumbing and ensure that the import works in the cases we know have been problematic in the past.
So, in addition to the current test file, please create a new test file, one that only concerns itself with this kind of test.
Please check in all three scrivener files and write tests that verify that when we load each Scrivener file from the file system, it'll produce a known JSON object representing a Plottr file.

main/modules/files.js
```diff
@@ -420,6 +420,9 @@ function createRTFConversionFunction(sender) {
 }
 
 function createFromScrivener(importedPath, sender, isLoggedIntoPro) {
+  if (importedPath.error) {
```
Quiescent: @jeanawhou, is this correct?  How can a string have an `error` property?

Quiescent:
Please resolve the conflicts.
Note that many of the functions in conflict have moved to the socket server; unless you've made changes to the file functions (which I don't think you did), it's safe to delete them from `files.js` simply.

main/modules/files.js
jeanawhou: I think its from the reducer 😅 

This is what i logged from the console when i import a non-scrivener/wrong format like folder
`importedPath { type: 'IMPORT_ERROR', error: 'Wrong file format' }`

Quiescent: I think this is an error from the client before asking the server to import.
One sec while I confirm this suspicion.

Quiescent (approved):
Well done, @jeanawhou !  LGTM :)

PR #812, opened 2022-06-09
Port auto save to the socket server

Quiescent:
# Motivation
Auto-save (and backup) were the last two significant file-saving procedures still left in the main process.
We believe that locking on the main process might contribute to the freezing and saving problems that users report.

# Approach
This PR extends the API of the socket client to add facilities for saving, backing up, logging directly to renderers, and responding to events on the socket server.


jeanawhou (approved):
LGTM :)

cameronsutter (approved):

PR #816, opened 2022-06-13
Nested bullet points

Quiescent:
# Motivation
Bullet lists and numbered lists are only available on the root-level.
Our users would like them to be available at indented levels.

# Approach
This PR enables list items to have bullet/numbered lists as children.
It adds two buttons to the RCE toolbar to indent and decent a numbered section; the indentation controls are disabled when the cursor isn't in a list of some kind.
This PR also binds `tab` and `shift+tab` to indent and dedent respectively to mimic similar hotkeys in other editors.

https://user-images.githubusercontent.com/1457336/173395972-ec99249e-86ec-4126-8b93-d540ecb573be.mp4


jeanawhou (approved):
LGTM :)

PR #820, opened 2022-06-14
Fix offline resume

Quiescent:
# Motivation
Resuming from offline is broken (it turns out to be a permission problem).
Additionally, Plottr didn't propagate failed Promises from the Firebase worker correctly.
This PR aims to fix both problems.

# Approach
This PR adds new reply types for errors received from the worker.
It also instruments all remaining errors on the worker and catches promises that Plottr didn't handle.
The rules for updating the `editors` key weren't correct; this PR adds a rule to be able to modify it like other keys.
Finally, this PR adds a new facility to retry effects, see `effect.js`.


jeanawhou (approved):
LGTM :)

PR #824, opened 2022-06-17
Bugfix/import scrivener

jeanawhou:
Ticket: https://airtable.com/appfbr7vlSqIfZ9Sn/tblAmQkxs32m46J4N/viwDW5z6rnAo0iXmL/recoWiPcB3BXX6Ofm?blocks=hide

@Quiescent @cameronsutter i've checked in sarmad's failing file and adjusted the importer.test.js to return errors


lib/plottr_import_export/src/importer/scrivener/importer.js
```diff
@@ -95,49 +95,53 @@ function ScrivenerImporter(filePath, isNewFile, state, convertRTFToSlate) {
...
+      return uuidsToIntegerIds({
+        ...newJson,
+        file: {
+          ...newJson.file,
+          lastOpened: new Date(),
+        },
+      })
+    })
+    .catch((error) => {
+      return new Error(error)
```
Quiescent, line 143: Please return the error directly instead.
(Otherwise, we'll lose the stack trace).

lib/plottr_import_export/src/importer/scrivener/importer.js
```diff
@@ -1160,29 +1179,33 @@ export function isNotExcludedAttribute(attribute) {
...
+              }
+            : null
+        }
+      })
+      .filter(Boolean)
 
-  return mergePairAttributes(flatAttributes)
+    return mergePairAttributes(flatAttributes)
+  } catch (error) {
+    return new Error(error)
```
Quiescent, line 1207: Ditto for this one.  If we construct a *new* error then the stack trace reflects this location instead of the original error.

Quiescent: Oh, nvm.  I was looking at an older commit.  Sorry! :)

lib/plottr_import_export/src/importer/scrivener/importer.js
Quiescent: Sorry.  I was wrong here.

Quiescent:
Good work, @jeanawhou 
Thank you for your diligence :+1:

PR #825, opened 2022-06-20
Normaliser dont allow images last

Quiescent:
# Motivation

Right now, if a user adds an image to the RCE, and it's the last element in the editor, they won't be able to add text after the image.
We'd like users to be able to have images as the last element and still add text afterwards.

# Approach

This PR adds a normaliser that ensures that an empty paragraph is added after images when they're the last element.


jeanawhou (approved):
LGTM :)

cameronsutter (approved):

PR #826, opened 2022-06-21
Backup offline backup when resuming

Quiescent:
# Motivation

Offline mode is a beta and could go wrong.
We want to reduce user impact by introducing a new mechanism for backing up files.

# Approach

Save backups off offline files with the prefix `_resume-backup_6-21-2022-13h21m48_` in front of the file name in the offline folder when we resume.
@cameronsutter, it doesn't clean up these files *ever*.
I'm not sure what the best approach is, but this is the safest, barring our users running out of HDD space.

# Other Fixes Included in this PR

 - Remove synchronous FS functions from the save offline backup function.
 - Move the Resume component to the root to avoid re-rendering and potentially messing up its process.
 - Log more context when the socket server receives a message it doesn't yet handle.
 - Fix the resume message when we accept what's on the cloud (i.e. we didn't change -- and it doesn't matter what happened on the cloud).
 - Only show the file name of offline files rather than the full path to the file.
 - Never open an offline file as the last known file (avoids modifying offline files directly while connected).


jeanawhou (approved):
LGTM :)

PR #827, opened 2022-06-23
added unit test for previous broken file; used the right json to match the relevant files of the scrivener project

jeanawhou:
@Quiescent @cameronsutter its broken because i forgot to add the unit test for the previously broken file that sarmad attached on the ticket (i also attached it here).

I reverted the first unit test to use the right json (example3) to match its relevant files.

The second unit test is for the Untitled.scriv that i attached

[Untitled.scriv.zip](https://github.com/cameronsutter/plottr_electron/files/8968264/Untitled.scriv.zip)


Quiescent (approved):
Great.  Thanks, @jeanawhou!
LGTM :)

PR #828, opened 2022-06-27
added reducer on delete book image

jeanawhou:
Ticket: https://airtable.com/appfbr7vlSqIfZ9Sn/tblP6AWIpvkOYZ0jR/viwou8K2UyUSkWGts/reckjk7DheZ43N5p1?blocks=hide

Demo: 

https://user-images.githubusercontent.com/19387007/175967311-b748d6f8-2226-4ff2-bed6-9ce338aeade3.mp4

### DIFF
--- a/lib/plottr_components/src/components/images/ImagePicker.js
+++ b/lib/plottr_components/src/components/images/ImagePicker.js
@@ -128,8 +128,15 @@ const ImagePickerConnector = (connector) => {
     }
 
     deleteImage = () => {
+      const { bookId, isBookImage, chooseImage, actions } = this.props
+
       if (this.state.deleting) {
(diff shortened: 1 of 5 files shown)


Quiescent (changes requested):
I don't think this is the way to do it.
There are two problems:
 1.  We're bleeding details about the component that uses the image picker into the image picker.  The image picker should never need to know about the components that use it.
 2.  If you delete the image from an image picker elsewhere in the app, then the problem still happens.

The real solution is to listen for DELETE_IMAGE in the books reducer and iterate through all of the books to find which ones use it.  It would be best if you made the `imageId` null for *all* of the books that use it.

We should probably also check that characters, notes and places don't have the same problem.

PR #830, opened 2022-06-30
Bugfix/delete book image

jeanawhou:
Ticket: https://airtable.com/appfbr7vlSqIfZ9Sn/tblP6AWIpvkOYZ0jR/viwou8K2UyUSkWGts/reckjk7DheZ43N5p1?blocks=hide

Demo:

https://user-images.githubusercontent.com/19387007/176675796-90488102-a84f-4368-a527-828a76f16ba0.mp4

### DIFF
--- a/lib/plottr_components/src/components/images/ImagePicker.js
+++ b/lib/plottr_components/src/components/images/ImagePicker.js
@@ -128,11 +128,8 @@ const ImagePickerConnector = (connector) => {
     }
 
     deleteImage = () => {
-      const { chooseImage, actions } = this.props
-
       if (this.state.deleting) {
-        actions.deleteImage(this.state.selectedId)
-        chooseImage(null)
+        this.props.actions.deleteImage(this.state.selectedId)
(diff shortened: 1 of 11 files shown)


lib/plottr_components/src/components/images/ImagePicker.js
```diff
@@ -128,8 +128,11 @@ const ImagePickerConnector = (connector) => {
     }
 
     deleteImage = () => {
+      const { chooseImage, actions } = this.props
+
       if (this.state.deleting) {
-        this.props.actions.deleteImage(this.state.selectedId)
+        actions.deleteImage(this.state.selectedId)
+        chooseImage(null)
```
Quiescent: This part doesn’t make sense to me.  You should be able to delete an image in the picker without choosing no image.  What if you want to delete a bunch of images and then pick an image?

Quiescent (changes requested):
Looks great.  Just had one comment about choosing null when we delete.

Quiescent (approved):
Great work.  LGTM!

PR #831, opened 2022-07-04
Upgrade docx

Quiescent:
# Motivation
We've had problems exporting to Word for a number of releases.
Some of them are completely mysterious:
 - we've extracted the word doc and checked the XML and it seems fine,
 - other programs can still open files exported by Plottr, and
 - images are sometimes exported and sometimes not.
We believe that some of these problems relate to schema changes that we haven't received because we used an older version of the `docx` package.

# Approach
In this PR, I've upgraded `docx` and made only one package depend on it: `plottr_import_export`.
The new package eschews side-effects in favour of constructing the `docx` tree immutably (nice!).
In my testing, the files that this branch exports are at least as good as they were before; I can open them in various programs.
Only Google docs seems to recognise the images.

I'm proposing that we merge this change because it's net positive in the sense that it both simplifies the build and upgrades docx.
Once merged, we can test what the files look like with a proper version of MS Word (I only have open source alternatives at my disposal).


lib/plottr_import_export/src/exporter/word/exporters/places.js
```diff
@@ -18,48 +17,50 @@ export default function exportPlaces(state, doc, options) {
...
         )
-        paragraphs.push(new Paragraph({ children: [image] }))
       }
     }
     if (options.places.descriptionHeading) {
       paragraphs.push(new Paragraph({ text: t('Description'), heading: HeadingLevel.HEADING_3 }))
     }
     if (options.places.description) {
-      paragraphs.push(new Paragraph(pl.description))
+      paragraphs.push(new Paragraph({ text: pl }.description))
```
jeanawhou: @Quiescent im not sure if this is intentional but i haven't encountered a syntax like this `{ text: pl }.description)` :sweat_smile:

Quiescent: Oh no!  That's not intentional at all 😂 
Thanks for spotting it!

cameronsutter (approved):
Let's merge it so we can get some testing done around Word

PR #835, opened 2022-07-07
Sketch out what the `PlotlineModal` might look like

Quiescent:
# Motivation

Plottr can't apply templates or move plot lines when the act structure is enabled.
The problem is that we need to decide whether to increase the levels and beats in the destination and what the source beats should map to in the destination.
This PR adds a modal to help the user decide what to do.

https://user-images.githubusercontent.com/1457336/177751331-6ede7442-0e53-4fa3-919e-24edff579f48.mp4


Quiescent:
We've opted for a different design.

PR #861, opened 2022-08-18
Socket server busy

Quiescent:
# Motivation
It's hard to know whether it's safe to close Plottr, and users have complete control over the Plottr process when it's a desktop app.
This has led users to close the app while it's busy saving in the past.

# Approach
Register various tasks on the socket server as important and do two things with that information:
 1. Don't let the user refresh or quit while the socket server is busy with those tasks.
 2. Indicate that Plottr is busy with a pulsating icon.

# Demo

https://user-images.githubusercontent.com/1457336/185377387-4288a8da-96a1-41b5-8a4b-a3e1a8b1cbff.mp4


jeanawhou (approved):
LGTM :)

PR #862, opened 2022-08-22
Tweak recent files menu

jeanawhou:
Ticket: https://airtable.com/appfbr7vlSqIfZ9Sn/tblAmQkxs32m46J4N/viwURSVZRtP7uz0I8/recbb23v05y0o07IH?blocks=hide

Demo: 

https://user-images.githubusercontent.com/19387007/185826225-5437c523-ce00-43e0-83ac-e9c653b0d446.mp4


lib/plottr_components/src/components/dashboard/files/FileActions.js
```diff
@@ -108,24 +120,28 @@ const FileActionsConnector = (connector) => {
...
+          </Button>
+          <Dropdown id={`file-action-${id}`} onSelect={doTheThing}>
+            <Dropdown.Toggle noCaret>
+              <Glyphicon glyph="option-horizontal" />
+            </Dropdown.Toggle>
+            <Dropdown.Menu>
+              {isCloudFile || osIsUnknown || missing ? null : (
+                <MenuItem eventKey="show">{showInMessage}</MenuItem>
+              )}
               <MenuItem eventKey="remove">{t('Remove from this list')}</MenuItem>
```
Quiescent: This shouldn't be visible on web, i.e. when `osIsUnknown`.

lib/plottr_components/src/components/dashboard/files/FileActions.js
```diff
@@ -108,24 +120,28 @@ const FileActionsConnector = (connector) => {
...
-            {isCloudFile || osIsUnknown || missing ? null : (
-              <MenuItem eventKey="show">{showInMessage}</MenuItem>
-            )}
-            {missing || offline ? null : <MenuItem eventKey="rename">{t('Rename')}</MenuItem>}
-            {(isCloudFile && permission !== 'owner') || missing ? null : (
-              <MenuItem eventKey="delete">{t('Delete')}</MenuItem>
-            )}
-            {(isTemp || missing) && !osIsUnknown && (
+        <ButtonGroup>
+          <Button bsSize="small" onClick={handleOpen}>
```
Quiescent: If the file's missing, we shouldn't show any of these three buttons, or they should be disabled.

Quiescent:
This change is great!  It looks way better than what we had.
I want to propose another design tweak for a future release: change file renaming for web so that it turns the name in the file list into an input box instead of bringing up the input popup.

lib/plottr_components/src/components/dashboard/files/FileActions.js
jeanawhou: Good catch @Quiescent thank you for checking :bow:

jeanawhou:
> This change is great! It looks way better than what we had. I want to propose another design tweak for a future release: change file renaming for web so that it turns the name in the file list into an input box instead of bringing up the input popup.

Should this be my next task @Quiescent ? :sweat_smile:

Quiescent:
> > This change is great! It looks way better than what we had. I want to propose another design tweak for a future release: change file renaming for web so that it turns the name in the file list into an input box instead of bringing up the input popup.
> 
> Should this be my next task @Quiescent ? sweat_smile

No.  We have a bunch of tasks on the board.  If we decide to do this (@cameronsutter FYI) then it'll be for the next release unless Cameron decides to bring it into this one.  I'll go through the tasks on the board and tag you on what I think should be next :)

jeanawhou:
> > > This change is great! It looks way better than what we had. I want to propose another design tweak for a future release: change file renaming for web so that it turns the name in the file list into an input box instead of bringing up the input popup.
> > 
> > 
> > Should this be my next task @Quiescent ? sweat_smile
> 
> No. We have a bunch of tasks on the board. If we decide to do this (@cameronsutter FYI) then it'll be for the next release unless Cameron decides to bring it into this one. I'll go through the tasks on the board and tag you on what I think should be next :)

thank you so much :bow:  PR updated :sweat_smile:

Quiescent (approved):
Looks good.  Please merge these with web and check whether the changes work as expected there.
LGTM! :)

PR #866, opened 2022-09-01
Firebase upgrade

Quiescent:
# Motivation
The latest version of Firebase offers many advantages over the current version.
For one thing, we occasionally run into a deadlock-like situation in version 8; that deadlock is not in version 9 (according to Google developers).


lib/plottr_firebase/src/api.js
```diff
@@ -387,78 +374,68 @@ const api = (auth, database, storage, baseAPIDomain, development, log, isDesktop
...
+          }))
           authorisedDocuments.push(document)
         })
-        return Promise.all(authorisedDocuments).then((documents) => {
-          return documents.map((document) => {
-            return {
-              ...document,
-              isCloudFile: true,
-            }
+        Promise.all(authorisedDocuments)
```
jeanawhou, line 390: @Quiescent is it ok not to return this promise?

jeanawhou:
@Quiescent i just have a question but otherwise it looks good to me :)

lib/plottr_firebase/src/api.js
Quiescent: Yup.  I intended not to return that Promise.  That function expects a void return because we're registering a callback (that's the API that Google gave us; they did that because we're subscribing to a store, not getting a single value).

PR #867, opened 2022-09-02
Bugfix/characters tab

jeanawhou:
Tickets: 
https://airtable.com/appfbr7vlSqIfZ9Sn/tblAmQkxs32m46J4N/viwDW5z6rnAo0iXmL/reczmXBuxlJCOtZEI?blocks=hide
https://airtable.com/appfbr7vlSqIfZ9Sn/tblAmQkxs32m46J4N/viwDW5z6rnAo0iXmL/rec2GU04R41w7ha6G?blocks=hide
https://airtable.com/appfbr7vlSqIfZ9Sn/tblAmQkxs32m46J4N/viwDW5z6rnAo0iXmL/reclJSdbyxMDX0PKi?blocks=hide

Demo

https://user-images.githubusercontent.com/19387007/188116119-f6721669-aced-48ad-ae72-f79a8d26f117.mp4

### DIFF
--- a/lib/plottr_components/src/components/characters/CharacterListView.js
+++ b/lib/plottr_components/src/components/characters/CharacterListView.js
@@ -305,23 +305,6 @@ const CharacterListViewConnector = (connector) => {
     const renderVisibleCharacters = (categoryId) => {
       if (!visibleCharactersByCategory[categoryId]) return []
 
-      if (!!attributeTabId && attributeTabId !== 'all') {
-        const filterByBookCharactersByCategory = visibleCharactersByCategory[categoryId].filter(
-          (ch) => ch.bookIds.includes(attributeTabId)
-        )
-
-        return filterByBookCharactersByCategory.map((ch) => (
-          <CharacterItem
-            key={ch.id}
-            character={ch}
-            selected={ch.id == selectedCharacteId}
-            startEdit={editSelected}
-            stopEdit={stopEditing}
-            select={() => uiActions.selectCharacter(ch.id)}
-          />
-        ))
-      }
-
       return visibleCharactersByCategory[categoryId].map((ch) => (
         <CharacterItem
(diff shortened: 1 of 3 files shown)


cameronsutter:
The demo LGTM

lib/plottr_components/src/components/characters/CharacterListView.js
```diff
@@ -305,6 +305,23 @@ const CharacterListViewConnector = (connector) => {
     const renderVisibleCharacters = (categoryId) => {
       if (!visibleCharactersByCategory[categoryId]) return []
 
+      if (!!attributeTabId && attributeTabId !== 'all') {
```
Quiescent: This needs to be a selector.
If the selector that we're using doesn't have a name that implies that it has this "filtering by book" functionality, then please add another selector.

Quiescent:
@jeanawhou changes look great.
Could you please move the logic for filtering characters out of the component?

Context: we haven't been strict enough about this in the past; it makes it difficult to maintain the code, and the mobile team needs to replicate behaviour each time we do it.

lib/plottr_components/src/components/characters/CharacterListView.js
jeanawhou: @Quiescent got it thank you! 🙇‍♀️

jeanawhou:
> @jeanawhou changes look great. Could you please move the logic for filtering characters out of the component?
> 
> Context: we haven't been strict enough about this in the past; it makes it difficult to maintain the code, and the mobile team needs to replicate behaviour each time we do it.

@Quiescent I see got it thank you for checking 🙇‍♀️

Quiescent:
Excellent work with that last commit, @jeanawhou!

Quiescent (approved):

PR #869, opened 2022-09-21
Improve known files

Quiescent:
# Motivation
Knowing how to open a file is guided by organically grown (™️ ) heuristics.
The logic is like nature, vibrant, and challenging to refactor 😉 
We think a better situation exists, where a concrete set of rules accurately governs behaviour.

# Approach
This PR aims to make file management synthetic and have everything line up in straight lines (as it should!) by:
 - using a new protocol scheme to locate files and decide how they should be saved and loaded, and
 - loading the list of available files for Pro from the authorisation collection directly.


jeanawhou (approved):
LGTM :)

PR #872, opened 2022-09-27
Lastopenedfile to socketserver

jeanawhou:
ticket: https://airtable.com/appfbr7vlSqIfZ9Sn/tblAmQkxs32m46J4N/viwURSVZRtP7uz0I8/recOFjSPHxDL5hm0Y?blocks=hide


main/listeners.js
```diff
@@ -34,42 +34,28 @@ import {
...
-              'state-fetched',
-              fileURL,
-              flags,
-              numberOfWindows(),
-              win.fileURL,
-              processSwitches.serialise()
-            )
-          })
-        }
+  ipcMain.on('pls-fetch-state', function (event, id, lastFile, proMode) {
```
Quiescent: We should instead call out to the socket server from here.

main/modules/lastOpened.js
```diff
@@ -1,48 +1,15 @@
...
-      return lastFileURL
-    })
-    .catch((error) => {
-      if (error.code === 'ENOENT') {
-        return null
-      }
-      log.error('Error finding last opened file.  Refusing to open it.', error)
-      return null
-    })
+const lastOpenedFile = () => {
```
Quiescent, line 3: Use this method from the `pls-fetch-state` listener.

main/server/server.js
```diff
@@ -844,6 +848,22 @@ const setupListeners = (port, userDataPath) => {
               () => 'Error while getting current backups'
             )
           }
+          case LAST_OPENED_FILE: {
+            return handlePromise(
+              () => 'Getting the last opened file',
+              () => lastOpenedFile,
```
Quiescent: You need to call `lastOpenedFile` or unwrap the lambda here.

src/app/containers/Main.js
```diff
@@ -206,7 +206,7 @@ const Main = ({
       ipcRenderer.removeListener('state-fetched', stateFetchedListener)
     }
     ipcRenderer.on('state-fetched', stateFetchedListener)
-    ipcRenderer.send('pls-fetch-state', win.id, isInProMode)
+    ipcRenderer.send('pls-fetch-last-file', win.id, isInProMode)
```
Quiescent: This is what's preventing the file from opening.  We need the reply to `pls-fetch-state` to open a file :)

Quiescent: Oh wait, I see what you did later.  Reading further.

Quiescent: When you send on the renderer, it'll be heard on the main process.  It looks like you setup the listener for this ipc call in the renderer?

src/app/index.js
```diff
@@ -307,6 +307,28 @@ ipcRenderer.on('redo', (event) => {
   store.dispatch(ActionCreators.redo())
 })
 
+ipcRenderer.on('pls-fetch-last-file', (event, id, isInProMode) => {
```
Quiescent: This listens for messages from the main process.

Quiescent: I think that you can avoid calling `pls-fetch-last-file` and then `pls-fetch-state` if you put the call to fetch the last opened file in the main process's listener.  It's fine to use the socket server client from the main process.

main/server/server.js
```diff
@@ -844,6 +848,22 @@ const setupListeners = (port, userDataPath) => {
               () => 'Error while getting current backups'
             )
           }
+          case LAST_OPENED_FILE: {
+            return handlePromise(
+              () => 'Getting the last opened file',
+              () => lastOpenedFile(),
```
Quiescent: This task might take time, we should register it with the `statusManager`.
(Tasks registered with the `statusManager` will inform the renderer that the socket server is busy and also prevent Plottr from quitting.)

Quiescent (approved):
I'd like one small change before you merge (to use the `statusManager` in one more place).
But other than that, it's looking great.  Well done! 👍

jeanawhou:
> I'd like one small change before you merge (to use the `statusManager` in one more place). But other than that, it's looking great. Well done! 👍

@Quiescent thank you so much for your teaching, guiding and advising me until i finish this task 🥲 🙇‍♀️

PR #873, opened 2022-09-27
Update db scripts with auditing etc

Quiescent:
Augment the two new db scripts with auditing and dry run capabilities in preparation for the production run.

### DIFF
--- a/lib/plottr_firebase/migrations/copy-known-file-information-to-authorisations.js
+++ b/lib/plottr_firebase/migrations/copy-known-file-information-to-authorisations.js
@@ -1,7 +1,9 @@
-const fs = require('fs')
 const admin = require('firebase-admin')
 const readline = require('node:readline')
 const { stdin, stdout } = require('node:process')
+
+const { logChange } = require('./log-change')
+const { sequencePromises } = require('./util')
 
 if (!admin.apps.length) {
   if (!process.env.FIREBASE_ENV || process.env.FIREBASE_ENV === '') {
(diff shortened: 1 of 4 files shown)


jeanawhou (approved):
LGTM :)

lib/plottr_firebase/migrations/copy-known-file-information-to-authorisations.js
```diff
@@ -23,14 +25,92 @@ if (!admin.apps.length) {
...
+            console.log('User exists.  Executing script.')
+            const userIds = userId ? Promise.resolve([userId]) : allUserIds()
+            userIds
+              .then((uids) => {
+                console.log(`Will run script for users with ids: ${JSON.stringify(uids, null, 2)}`)
+                return sequencePromises(
+                  uids.map((uid) => {
+                    return () => {
+                      console.log(`> ${uid}`)
+                      return markDeletedAuthorisedDocumentsAsDeleted(
```
cameronsutter, line 56: I think this migration is doing the wrong thing. The name of the file suggests it should be the  copy-known-file-info-to-authorisations one

lib/plottr_firebase/migrations/mark-deleted-authorised-documents.js
```diff
@@ -60,26 +142,32 @@ const deletedFiles = (userId) => {
...
+  })
 }
 
-const markDeletedAuthorisedDocumentsAsDeleted = (userId) => {
+const markDeletedAuthorisedDocumentsAsDeleted = (userId, executingUserId, readOnly) => {
   const database = admin.firestore()
+  const runTransactionWithAuditing = logChange(readOnly)(
+    database,
+    executingUserId,
+    'copy-known-file-information-to-authorisations.js',
```
cameronsutter, line 168: i think this line should be something to do with this migration, right? Something along the lines of 'mark-deleted-authorised-documents-as-deleted`

lib/plottr_firebase/migrations/copy-known-file-information-to-authorisations.js
Quiescent: Oops.  Copy-paste error.
Only the name is wrong :sweat_smile:

lib/plottr_firebase/migrations/mark-deleted-authorised-documents.js
Quiescent: Same thing here... thanks for spotting these! :)

PR #875, opened 2022-09-29
Attributes per book

Quiescent:
- Make all character attributes (custom, standard and template) appear for all characters despite the book.
 - Make all character attribute values distinct per book.
 - Hide book tabs until any character is associated to a book.


jeanawhou (approved):
LGTM :) i think we can create a new PR for the additional ui changes that cameron mentioned on the #dev-desktop channel 😅

PR #877, opened 2022-10-12
Export new character attributes, disable busy logo & reduce pressure saving templates

Quiescent:
# Export new character attributes
Plottr now exports character attributes per book.
It does this using the character selectors that are sensitive to book selections:
 - `characterAttributesSelector`, and
 - `allDisplayedCharactersSelector`.
(The feature added more selectors that are sensitive to the currently selected book, but these are what was needed to make export work.)

# Disable busy logo
We've decided to disable the busy indicator until a subsequent release because we didn't have time to consider how it should look and whether we should have it.
The socket server still tracks whether we're busy and prevents the system from shutting down if we are.

# Reduce pressure on template store
When templates update, the server previously fetched them and tried to update their corresponding records in the store in parallel.
This caused a lot of pressure on the host OS's file system; Windows couldn't take the pressure and regularly truncated the file early.
In this PR, we first fetch all the templates, compute the new store object, and write it all to the store in one operation.
This should greatly reduce the pressure on the file system and reduce the likelihood of errors making their way to support.

### DIFF
--- a/lib/plottr_import_export/src/exporter/scrivener/v2/exporters/__tests__/characters.test.js
+++ b/lib/plottr_import_export/src/exporter/scrivener/v2/exporters/__tests__/characters.test.js
@@ -5,10 +5,10 @@ import { headingTwo, paragraph } from 'components/__fixtures__'
 import default_config from '../../../../default_config'
 
 describe('exportCharacters', () => {
-  let documentContents = {}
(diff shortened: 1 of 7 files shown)


jeanawhou (approved):
LGTM :)

cameronsutter (approved):
LGTM

PR #890, opened 2022-11-03
Enhancement/card dialog

jeanawhou:
Ticket: https://airtable.com/appfbr7vlSqIfZ9Sn/tblP6AWIpvkOYZ0jR/viwou8K2UyUSkWGts/recaLaR3Mh0iXTO1c?blocks=hide

Test Result:

image


Quiescent:
You have an unused connected component at the top of `Card.js`: `  const CardDialog = UnconnectedCardDialog(connector)`; please delete it.

lib/pltr/v2/actions/__tests__/ui.test.js
```diff
@@ -0,0 +1,125 @@
...
+    })
+
+    it('should have false as initial valuefor isOpen', () => {
+      expect(isOpen).toBeFalsy()
+    })
+  })
+
+  describe('given cardDialog actions are dispatched', () => {
+    describe('given setCardDialogOpen is dispatched', () => {
+      store.dispatch(setCardDialogOpen(1, 1, 1))
```
Quiescent: Does a card with `id` 1 exist?
I think `pltr` should prevent the card from being opened if it doesn't exist.
You can handle that in either the main reducer (`lib/pltr/v2/reducers/main.js`) or by using `redux-thunk` (see `editFileName` in `actions/ui.js`.)
In this case, I think the appropriate choice is probably to change your action because an action should never be dispatched if the card doesn't exist.

Quiescent: If you did all that, then you'd need to change the expectations of this test and add another test where we first add a card before manipulating it.

lib/pltr/v2/actions/ui.js
```diff
@@ -189,3 +191,11 @@ export function selectCharacterAttributeBookTab(bookId) {
 export function selectCharacter(id) {
   return { type: SELECT_CHARACTER, id }
 }
+
+export function setCardDialogOpen(cardId, beatId, lineId) {
+  return { type: SET_CARD_DIALOG_OPEN, cardId, lineId, beatId }
```
Quiescent: This is where you *might* use `redux-thunk` to check whether the card, beat and line actually exist before dispatching the action.

Quiescent:
This PR looks great!  Good job :slightly_smiling_face: 

I only have one comment about state invariants (i.e. that we don't say that a dialogue is open for a non-existent card/beat/line) that I think is worth dealing with before we merge this.

Quiescent (approved):
Looks good.  LGTM! :+1:

PR #891, opened 2022-11-07
Better autosave

Quiescent:
# Motivation
Auto-save is an error-prone component of Plottr Desktop.  Over time, we've seen various errors from host machines.  These include (for example) permission errors and partial writes of files.

Over time, Plottr became more pedantic about successfully saving a file.  Unfortunately, we've seen this pedantic change cause problems.

One of the problems we've identified is when a high frequency of changes to the store causes pressure on the auto-save component; sometimes, Plottr places increased pressure on the OS's file system to save files.  Some OSs (and indeed hardware) handle this better than others (looking at you, Windows!), and we can't rely on the strength of the guarantee from an operating system to fulfil its persistence guarantees.

# Approach
This PR introduces a `Saver` component that doesn't operate via redux middleware.  Plottr initialises a `Saver` when it boots a file, and the `Saver` determines when to:
 - save,
 - record an offline file (for the offline mode in Pro), and
 - back up a file (including backing up Pro files on the local disk).
 
The `Saver` tracks the last saved state and maintains a timer to save and back up on a fixed interval.  It launches an auto-save job every ten seconds and a backup job every sixty seconds.  If the last save job is yet to complete, it enqueues the next auto-save it'll execute when the previous save job completes.  It behaves similarly for backup jobs.

The `Saver` calls the new `saveFile` and `backupFile` functions (in `src/app/save.js`) to:
 - check whether we ought to be able to save,
 - check whether we should instead save an offline backup file (in Pro with offline mode enabled), and
 - whether we should be able to back up.

This PR adds extensive tests to verify the job scheduling behaviour of `Saver` and the logic encoded by `save.js`.  It also removes the pedantic checks the socket server performed.  The socket server now checks whether files can be saved to disk, opens a file handle, writes the file, waits for the operation to flush from the OS's FS operation queue and then responds to the request.


jeanawhou (approved):
LGTM :)

lib/pltr/v2/selectors/project.js
```diff
@@ -29,3 +37,61 @@ export const fileURLSelector = (state) => state.project.fileURL
...
+  isResumingSelector,
+  isOfflineSelector,
+  offlineModeEnabledSelector,
+  hasAllKeysSelector,
+  isCloudFileSelector,
+  (fileURL, isResuming, isOffline, offlineModeEnabled, hasAllKeys, isCloudFile) => {
+    return (
+      !!fileURL &&
+      !isResuming &&
+      (!isCloudFile || !(isCloudFile && isOffline && !offlineModeEnabled)) &&
```
cameronsutter, line 56: this line is complex. I want to make sure I'm understanding it.

not a cloud file OR an offline cloud file with offline mode … off? That feels like the opposite of what we want.

i'm not sure i'm following the ! before the parens here (`!(isCloudFile…`. Is that to negate the `!offlineModeEnabled` so that it makes the whole statement:

not a cloud file OR an offline cloud file with offline mode on

?

src/app/save.js
```diff
@@ -0,0 +1,68 @@
+import { difference } from 'lodash'
+
+import { helpers, selectors, SYSTEM_REDUCER_KEYS, emptyFile } from 'pltr/v2'
+
+const emptyFileState = emptyFile('DummyFile', '2022.11.2')
```
cameronsutter, line 5: Are we going to have to change this version number every release?

cameronsutter (commented):
So much deleted code :)

lib/pltr/v2/selectors/project.js
Quiescent: Hmmm, it does look complicated.

Here's my reasoning:
The entire expression evaluates to `true` if we *can* save (where "save" means save to disk because the concept doesn't exist on Firebase.)
Generally, you can't save a cloud file, but you are allowed to when it's a cloud file and `offlineMode` is enabled.
Later, in `save.js`, we check whether we should save an offline file.
If we're allowed to save, *but* we can't save an offline file, and we're a pro file, then `save` resolves *without* saving.

I might've made a mistake here, so let's apply DeMorgan's theorem:
```
!isCloudFile || !(isCloudFile && isOffline && !offlineModeEnabled)
= !isCloudFile || (!isCloudFile || !isOffline || offlineModeEnabled)
= !isCloudFile || !isCloudFile || !isOffline || offlineModeEnabled
= !isCloudFile || !isOffline || offlineModeEnabled
```
So it ends up being either:
 - we're not a cloud file, or
 - we're not offline, or
 - offline mode is enabled.

I think the offline mode check *might* be redundant here, but I like double-checking it in case the logic for checking whether to save an offline file instead of an ordinary file changes.

Does it make sense to you?
Should we apply that simplification I made to reason about the logic?

src/app/save.js
Quiescent: Nope.  I just needed the keys of an initial file.

lib/pltr/v2/selectors/project.js
cameronsutter: ok cool. No need to simplify. That makes sense to me now. Thanks!

cameronsutter:
LGTM

PR #899, opened 2022-11-10
Download images into pro local backups

Quiescent:
- Re-instate the firebase backup mechanism.
 - Ensure that local backups of pro files have the correct name.
 - Always backup a self-contained version of the plottr file (i.e. download all `storage://` protocol items back into the file before saving it.)

### DIFF
--- a/src/app/__tests__/save.test.js
+++ b/src/app/__tests__/save.test.js
@@ -557,6 +557,8 @@ describe('saveFile', () => {
   })
 })
 
+const DUMMY_DOWNLOAD_IMAGE_FROM_STORAGE = () => {}
+
 describe('backupFile', () => {
   describe('given a whenClientIsReady that produces a dummy saveBackup', () => {
     describe('and a file that lacks the necessary keys', () => {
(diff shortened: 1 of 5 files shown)


jeanawhou (approved):
LGTM :)

cameronsutter (commented):
I could be reading this wrong, but my biggest concern here is that we're downloading the images for every backup, right?

Can we cache them locally somehow? Those aren't going to change often … except when they do of course

Quiescent:
> I could be reading this wrong, but my biggest concern here is that we're downloading the images for every backup, right?
> 
> Can we cache them locally somehow? Those aren't going to change often … except when they do of course

You're right.  This is why we didn't have this feature in the first place. :thinking: 
I think it's safe to have a per-session in-memory cache.
I'll start work on that now.
We can always change it to a disk-level cache if we have to, but in-memory is easier, to begin with.

Quiescent:
> > I could be reading this wrong, but my biggest concern here is that we're downloading the images for every backup, right?
> > Can we cache them locally somehow? Those aren't going to change often … except when they do of course
> 
> You're right. This is why we didn't have this feature in the first place. thinking I think it's safe to have a per-session in-memory cache. I'll start work on that now. We can always change it to a disk-level cache if we have to, but in-memory is easier, to begin with.

@cameronsutter , I've implemented a cached version.  WDYT?

cameronsutter (approved):
LGTM

PR #902, opened 2022-11-15
Handle UI per user on pro

Quiescent:
# Motivation
Pro doesn't track changes to the `ui` key because that would lead to conflicts when collaborating with others.
*Who should write into `ui` when three users modify the same document?*

We want each user to be able to track their own `ui` across sessions.
We'd also like to communicate where in the UI each user is so that we can build features like indicators on the UI that show where your collaborators are (TBD.)

# Implementation

## Schema Change
Plottr now stores an additional key on the root of the `ui`: `collaborators`.
`collaborators` is an object with the following schema:

```json
{
  ...uiState,
  "collaborators": {
    "collaborators": [
      {
       ...uiState without "collaborators",
        "id": "uid of logged-in collaborator",
      },
      ...
    ],
    "viewers": [
      {
        ...uiState without "collaborators",
        "id": "uid of logged-in viewer",
      },
      ...
    ]
  }
}
```

The owner modifies the root of the `ui` as though it's an ordinary Plottr file.
Each collaborator modifies a record in either "collaborators" or "viewers", depending on their permission on the file.
The collaborator `ui` state is identical to the root `ui` state except that it doesn't include the key `collaborators`.

## Reducer Change
Plottr now attaches the currently logged-in user id, permission and whether the user *should* be logged in to every action.
The `ui` reducer reads these three data to decide whether or not to write to `ui` and where in `ui` it should write to.

## Selector Change
When selecting *any* `ui` state, the `uiSelector` now selects the appropriate sub-ui state for the matching logged-in user.


jeanawhou (approved):
LGTM :)

PR #903, opened 2022-11-15
Export all characters to scrivener when no tabs

Quiescent:
- Export all characters to Scrivener when the book tabs aren't exported.


jeanawhou (approved):
LGTM :)

PR #904, opened 2022-11-15
Enhancement/card dialog

jeanawhou:
Couple of bugs @Quiescent found 😅 🙇 

1. Now that the open card is tracked by redux we need to close the card when it's deleted.
2. When you change the chapter/plotline that a card belongs to from the card dialogue, it doesn't reflect the update in the dropdown controls.

Demo:

https://user-images.githubusercontent.com/19387007/201969591-804c6268-5c87-40a1-bf99-f335a00397ad.mp4


Quiescent (changes requested):
Looks great.  Please add tests for these actions.

Quiescent (approved):
Good work!  LGTM! :)

PR #910, opened 2022-11-21
Enhancement/book dialog

jeanawhou:
Tickets:
https://airtable.com/appqY1Wa4HybeLzHs/tblTBZjhv1VH6kFPH/viw5M5GJxLKIzJq9j/recPCq4CpHH9bh9A1?blocks=hide

https://airtable.com/appqY1Wa4HybeLzHs/tblTBZjhv1VH6kFPH/viw5M5GJxLKIzJq9j/recBCeX1lFC3V9wJt?blocks=hide

https://airtable.com/appqY1Wa4HybeLzHs/tblTBZjhv1VH6kFPH/viw5M5GJxLKIzJq9j/recsHMhmnc5cWY5pq?blocks=hide

Demo

https://user-images.githubusercontent.com/19387007/202980421-a99949e9-ad13-43cf-a072-6f773a099940.mp4


lib/plottr_components/src/components/project/BookDialog.js
```diff
@@ -24,38 +30,67 @@ const BookDialogConnector = (connector) => {
...
       return (
-        <Form horizontal>
+        <Form horizontal onKeyDown={this.handleDown}>
           <FormGroup>
             <Col componentClass={ControlLabel} sm={3}>
               {i18n('Book #')}
             </Col>
             <Col sm={8}>
-              <span className="lead">{this.props.bookNumber}</span>
+              {/* bookNumber will be fetched from the current `book`
```
Quiescent: But if you're creating a book, the number will be `books.length`, wont it?

lib/pltr/v2/store/initialState.js
```diff
@@ -96,6 +96,11 @@ export const uiState = {
     beatId: null,
   },
   collaborators: {},
+  bookDialog: {
+    isOpen: false,
+    bookId: null,
+    bookNumber: null,
```
Quiescent: I don't think you should have this piece of state here.
You can get the `bookNumber` from elsewhere (see previous comment).

lib/pltr/v2/actions/ui.js
```diff
@@ -213,6 +217,18 @@ export function setCardDialogClose() {
...
 
+export function setBookDialogOpen(bookId, bookNumber) {
+  return {
+    type: SET_BOOK_DIALOG_OPEN,
+    bookId,
+    bookNumber,
+  }
+}
+
+export function setBookDialogClose() {
```
Quiescent: Please rename this to `closeBookDialog`.

lib/pltr/v2/actions/ui.js
```diff
@@ -213,6 +217,18 @@ export function setCardDialogClose() {
   return { type: SET_CARD_DIALOG_CLOSE }
 }
 
+export function setBookDialogOpen(bookId, bookNumber) {
```
Quiescent: This interface is too generic.  I see you used it in two different ways previously.
That puts the onus on knowing how to use it on the component (and the writer of the component).
Please break it down into the following two functions instead:
 1. `openEditBookDialog(bookId)`, and
 2. `openNewBookDialog()`.
As discussed elsewhere, the book number isn't necessary and makes this confusing for the next person who wants to use this action.

lib/plottr_components/src/components/project/BookDialog.js
```diff
@@ -24,38 +30,67 @@ const BookDialogConnector = (connector) => {
...
       let title = this.titleRef.value
       let premise = this.premiseRef.value
       let genre = this.genreRef.value
       let theme = this.themeRef.value
-      this.props.actions.editBook(book.id, { title, premise, genre, theme })
-      this.props.cancel()
+      const newBookId = objectId(books.allIds)
+
+      if (!book) {
+        this.props.actions.addBook(newBookId, { title, premise, genre, theme })
```
Quiescent: Please defer as much logic as you can to `pltr`.
In this instance, you've computed the new book id in the component, but we could have done so in the reducer instead.
If there's no logic required to **use** `pltr`, then our components become cleaner and will likely have fewer bugs.

To that end, please:
 - remove `newBookId` from `addBook`, and
 - spread its other arguments out of that object.  i.e. please change the signature of add book to: `addBook(title, premise, genre, theme)`.  This will make it trivial for others to know what to supply.   If you keep it as an object, then whoever uses `pltr` has to research how it was used in other projects.
 
Finally, if possible, please change the signature of `editBook` to `editBook(bookId, title, premise, genre, theme)`.
If `editBook` is used in more places than just here, then please document the change in the readme of `pltr` in a new section like so:
```
# Breaking Changes
## <today's date>
 - `editBook(bookId, {title, premise, genre, theme}) -> editBook(bookId, title, premise, genre, theme)`
```

lib/plottr_components/src/components/project/BookDialog.js
```diff
@@ -64,11 +99,12 @@ const BookDialogConnector = (connector) => {
             </Col>
             <Col sm={8}>
```
Quiescent, line 98: (This is a long-standing problem): at some point, we need to change this into a controlled component.
Right now, it uses refs to grab the value when we save.
For these controls to work well with Pro, they need to update the stare with every key-stroke.

lib/plottr_components/src/components/project/BookDialog.js
```diff
@@ -24,38 +30,67 @@ const BookDialogConnector = (connector) => {
...
             <Col componentClass={ControlLabel} sm={3}>
               {i18n('Book #')}
             </Col>
             <Col sm={8}>
-              <span className="lead">{this.props.bookNumber}</span>
+              {/* bookNumber will be fetched from the current `book`
+                if `book` is not empty (editBook mode)
+                else bookDialogBookNumber (addBook mode)
+              */}
+              <span className="lead">{bookNumber || bookDialogBookNumber}</span>
```
Quiescent: Please write a `bookNumberSelector` to replace the `bookDialogBookNumberSelector` and remove `bookDialogBookNumber`.
The selector should produce the number of books in the project when there's no `bookId` for this dialog.

lib/plottr_components/src/components/project/BookList.js
```diff
@@ -145,6 +147,14 @@ const BookListConnector = (connector) => {
       })
     }
 
+    renderBookDialog = () => {
+      const { bookDialogBookId, bookDialogBookNumber, isBookDialogVisible } = this.props
+
+      if (!isBookDialogVisible) return false
+
+      return <BookDialog bookId={bookDialogBookId} bookNumber={bookDialogBookNumber} />
```
Quiescent: Please select the `bookDialogBookId` and `bookNumber` from the `BookDialog` rather than passing them down.

lib/pltr/v2/actions/__tests__/ui.test.js
```diff
@@ -35,13 +51,26 @@ const initialCardDialogState = {
   beatId: null,
 }
 
+const initialBookDialogState = {
+  isOpen: false,
+  bookId: null,
+  bookNumber: null,
```
Quiescent: I don't think the `bookNumber` is necessary.

lib/pltr/v2/actions/__tests__/ui.test.js
```diff
@@ -224,3 +253,153 @@ describe('cardDialog', () => {
...
+
+          const book = mapValues(allBooks, (value, key) => {
+            return value.title == exampleBookAttributes.title &&
+              value.premise == exampleBookAttributes.premise &&
+              value.genre == exampleBookAttributes.genre &&
+              value.theme == exampleBookAttributes.theme
+              ? true
+              : undefined
+          })
+          expect(Object.keys(book).length).toBeGreaterThan(0)
```
Quiescent: I don't think this test makes sense(?)
Surely, the length of keys in the book will still be 4?

lib/pltr/v2/actions/__tests__/ui.test.js
```diff
@@ -224,3 +253,153 @@ describe('cardDialog', () => {
...
+
+          const book = mapValues(allBooks, (value, key) => {
+            return value.title == newTitle &&
+              value.premise == currentBook.premise &&
+              value.genre == currentBook.genre &&
+              value.theme == currentBook.theme
+              ? true
+              : undefined
+          })
+          expect(Object.keys(book).length).toBeGreaterThan(0)
```
Quiescent: Again, I'm not sure that this test is good enough.  Perhaps break it into 4 assertions, one for each of the properties you're checking.

lib/pltr/v2/actions/books.js
```diff
@@ -8,8 +8,8 @@ import {
 } from '../constants/ActionTypes'
 import { book } from '../store/initialState'
 
-export function addBook() {
-  return { type: ADD_BOOK, book }
+export function addBook(newBookId, attributes) {
```
Quiescent: As discussed before, please don't supply the `newBookId` and spread `attributes` into four arguments.

lib/pltr/v2/actions/ui.js
```diff
@@ -1,3 +1,4 @@
+import { mapValues } from 'lodash'
```
Quiescent: I don't think you're using this(?)

lib/pltr/v2/reducers/books.js
```diff
@@ -42,6 +42,7 @@ const books =
           allIds: [...state.allIds, action.newBookId],
           [action.newBookId]: {
             ...action.book,
+            ...action.attributes,
```
Quiescent: You'll want to specify each attribute here and compute the `newBookId` just above the return statement.

lib/pltr/v2/selectors/ui.js
```diff
@@ -132,6 +136,17 @@ export const isCardDialogVisibleSelector = createSelector(cardDialogSelector, (c
   return cardDialog?.isOpen
 })
 
+export const bookDialogBookNumberSelector = createSelector(bookDialogSelector, (bookDialog) => {
```
Quiescent: This selector should be able to compute what the book number is based on whether there is a `bookId` (i.e. the book dialog is open, and how many books there are already.

Quiescent (changes requested):
This works, but it needs to defer more logic to `pltr`.
As it stands, you must compute a bunch of parameters from the components to call actions.
It should be as simple as possible for our components to use `pltr`.

lib/plottr_components/src/components/project/Book.js
```diff
@@ -30,16 +28,15 @@ const BookConnector = (connector) => {
...
       deleting: false,
       showTemplatePicker: false,
       showImagePicker: false,
     }
 
     chooseImage = (newId) => {
-      const id = newId == -1 ? null : newId
-      this.props.actions.editBook(this.props.book.id, { imageId: id })
+      const imageId = newId == -1 ? null : newId
+      this.props.actions.editBookImage(this.props.book.id, imageId)
```
Quiescent, line 39: :100: :tada:

lib/pltr/v2/actions/__tests__/ui.test.js
```diff
@@ -224,3 +250,178 @@ describe('cardDialog', () => {
...
+          const book = mapValues(allBooks, (value, key, obj) => {
+            if (typeof value == 'object' && !value.title) {
+              return value
+            }
+            return undefined
+          })
+          const filteredBook = Object.keys(book).filter((b) => !!book[b])
+          expect(filteredBook.title).toBeUndefined()
+          expect(filteredBook.theme).toBeUndefined()
+          expect(filteredBook.title).toBeUndefined()
```
Quiescent: I think this and the next line are typos(?)

lib/pltr/v2/actions/books.js
```diff
@@ -5,19 +5,26 @@ import {
...
+  return { type: ADD_BOOK, book, title, premise, genre, theme }
 }
 
 export function addBookFromTemplate(templateData) {
   return { type: ADD_BOOK_FROM_TEMPLATE, book, templateData }
 }
 
-export function editBook(id, attributes) {
-  return { type: EDIT_BOOK, id, attributes }
+//  Breaking Changes
```
Quiescent: Please document this change in the README.md file in `pltr`.
So at the end of the file at `lib/pltr/README.md`, add a new section that looks like this:
```
# Breaking Changes
## 2022-23-11
Old: `function editBook(id, attributes)`
New: `function editBook(id, title, premise, genre, theme)`

Quiescent (changes requested):
This is looking great!
Very well done, @jeanawhou :D 
I just have two minor comments for you to resolve before I merge it.

lib/pltr/v2/actions/__tests__/ui.test.js
jeanawhou: Great catch @Quiescent :bow:

Quiescent: Nearly there!  It still checks `title` three times :wink:

jeanawhou: @Quiescent sorry 🥲 thank you for your patience checking

Quiescent: No problem! :)

Quiescent (approved):
Great work.  LGTM!

PR #933, opened 2023-01-09
Restart socket server on disconnect

Quiescent:
# Motivation

In the latest version of Plottr, we've sand-boxed the renderer.
The renderer windows are now almost exactly the same as web browsers and much more secure.
Our libraries have changed a bit, though.
I discovered that the socket server now closes when the host computer goes to sleep.
This PR resolves that problem by rebooting the socket server when the connection exceeds ready-state 1, i.e. is closing or closed.

### DIFF
--- a/lib/plottr_components/src/components/timeline/TimelineWrapper.js
+++ b/lib/plottr_components/src/components/timeline/TimelineWrapper.js
@@ -33,12 +33,15 @@ import { checkDependencies } from '../checkDependencies'
 import { withEventTargetValue } from '../withEventTargetValue'
 import Scrollable from '../../utils/scrollable'
 import UnconnectedCardDialog from './CardDialog'
+import { makeMainProcessClient } from '../../../../../src/app/mainProcessClient'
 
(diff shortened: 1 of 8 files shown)


lib/plottr_components/src/components/timeline/TimelineWrapper.js
```diff
@@ -33,13 +33,16 @@ import { checkDependencies } from '../checkDependencies'
 import { withEventTargetValue } from '../withEventTargetValue'
 import Scrollable from '../../utils/scrollable'
 import UnconnectedCardDialog from './CardDialog'
+import { makeMainProcessClient } from '../../../../../src/app/mainProcessClient'
```
Quiescent: Erm.  this wasn't meant to be in the PR...

Quiescent: You didn't see anything 🐧 👋

jeanawhou (approved):
LGTM :)

PR #934, opened 2023-01-12
Bugfix/popover

jeanawhou:
This is the fix for the bug that ryan reported on dev-desktop
@cameronsutter @Quiescent the only case where i noticed that its still cutting the popover is when we hover a mid-position **card** but we can still scroll to view the entire popover content 😅 

Demo

https://user-images.githubusercontent.com/19387007/212068691-8377d30c-a7a3-423f-9170-fe2b2ecf6fd4.mp4

### DIFF
--- a/lib/plottr_components/src/components/PlottrFloater.js
+++ b/lib/plottr_components/src/components/PlottrFloater.js
@@ -1,6 +1,7 @@
 import React, { useCallback } from 'react'
 import PropTypes from 'prop-types'
 import { Popover, ArrowContainer } from 'react-tiny-popover'
+import cx from 'classnames'
 
 const PORTAL_ID = 'plottr-floater-portal'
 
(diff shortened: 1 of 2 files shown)


lib/plottr_components/src/components/PlottrFloater.js
```diff
@@ -35,7 +36,17 @@ const PlottrFloaterConnector = (connector) => {
             arrowColor={darkMode ? '#555' : 'white'}
             arrowSize={10}
             arrowStyle={{}}
-            className="popover-arrow-container"
+            className={cx('popover-arrow-container', {
+              resize:
```
Quiescent, line 40: So does this check whether the component is anchored to the middle of the window and occupies the other half of the window?

Quiescent (approved):
This is a very clever solution to the problem.  Well done 👍 
@cameronsutter check out the condition for resizing the popover 🙂

PR #935, opened 2023-01-17
Create desktop shortcut

jeanawhou:
While testing I noticed that the loading time on my pc is also too long

Demo

https://user-images.githubusercontent.com/19387007/212897402-341bd10f-226b-4577-99fc-e226a8876ef0.mp4


Quiescent (approved):
This is excellent work.
Well done, @jeanawhou !
LGTM!

PR #941, opened 2023-02-06
filter out empty non existing filter items; sorted filter items

jeanawhou:
@Quiescent @cameronsutter this is an enhancement for the filters on every tab. This includes

1. reuse the filter on every tab
2. hide the filter items that dont have matching content with the current view
3. sort each filter items including the custom attributes

i wasn't able to remove the book that is not used on any `card-item/card-view` for each tab from BookList on the `filter popover`. I get an error when i call the `currentViewSelector`, `allCharactersSelector` and other selectors inside the book selector.

The category list of the filter popover works ok on notes but not on character list. Not sure whats the problem.


Quiescent (approved):
Looks good.  LGTM!

PR #942, opened 2023-02-06
Pin plotlines

jeanawhou:


lib/pltr/v2/actions/lines.js
```diff
@@ -42,8 +46,55 @@ export function editLineColor(id, color) {
...
+
+    if ((!bothPinned && isPinnedPlotline) || (!bothPinned && isConflictWithPinned)) {
+      return false
+    }
+
+    const reorderedLines = reorderList(droppedPosition, originalPosition, lines)
+    return dispatch({ type: REORDER_LINES, lines: reorderedLines, bookId })
+  }
+
+export const togglePinPlotline = (line, lines, bookId) => (dispatch, getState) => {
```
Quiescent: Can we narrow this down so that you only supply the line?
You can select both the lines and the bookId from the state.

lib/pltr/v2/reducers/ui.js
```diff
@@ -640,6 +642,26 @@ const updateUI = (state, action) => {
       }
     }
 
+    case PIN_PLOTLINE: {
```
Quiescent, line 645: Can you please use `Math.max(0, <computed-number>)` and `Math.min(action.lines.length - 1)` here to prevent the count of pinned plot lines from exceeding the number of lines or going below zero.

Also, what happens when you change timelines?
Will the timeline have pinned lines when you get there?
We should probably reset this when the timeline changes.

lib/pltr/v2/selectors/ui.js
```diff
@@ -236,6 +241,43 @@ export const orientationSelector = createSelector(
...
+export const stickyHeaderCountSelector = createSelector(
+  pinnedPlotlinesSelector,
+  timelineViewIsStackedSelector,
+  hierarchyLevelCount,
+  selectedOrientationSelector,
+  (pinnedPlotlines, timelineViewIsStacked, hierarchyLevelCount, selectedOrientation) => {
+    if (selectedOrientation == 'vertical') {
+      return 1
+    }
+    if (timelineViewIsStacked) {
```
Quiescent: I'm not sure this is right, so on the stacked view we don't pin plot lines?

lib/pltr/v2/selectors/ui.js
```diff
@@ -236,6 +241,43 @@ export const orientationSelector = createSelector(
...
+
+export const stickyLeftColumnCountSelector = createSelector(
+  pinnedPlotlinesSelector,
+  timelineViewIsStackedSelector,
+  hierarchyLevelCount,
+  selectedOrientationSelector,
+  (pinnedPlotlines, timelineViewIsStacked, hierarchyLevelCount, selectedOrientation) => {
+    if (selectedOrientation == 'horizontal') {
+      return 1
+    } else if (timelineViewIsStacked) {
```
Quiescent: Same question here about the stacked view.

Quiescent: Oh never mind!
In stacked view, you can't flip the timeline so this will never be a problem.

lib/pltr/v2/selectors/ui.js
jeanawhou: @Quiescent im not sure what's the computation with stackedview.
It is `hierarchyLevelCount + pinnedPlotlines + 1`?

lib/pltr/v2/reducers/ui.js
jeanawhou: @Quiescent sorry im not familiar how we change the timeline 😅

Quiescent: The action is: `CHANGE_CURRENT_TIMELINE`.
You can reset the number of pinned lines there, but I actually think that'd be the wrong approach.
Better: you should an object per book to record the number of pinned timelines for each book.
i.e.:
```
{
   ...
   timeline: {
     ...
     pinnedPlotlines: {
       ...
       [bookId]: X,
     }
   }
}
```

Then you'd update the selector so that it selects from the book that's currently open.

lib/pltr/v2/actions/lines.js
```diff
@@ -42,8 +50,58 @@ export function editLineColor(id, color) {
   return { type: EDIT_LINE_COLOR, id, color }
 }
 
-export function reorderLines(lines, bookId) {
-  return { type: REORDER_LINES, lines, bookId }
+export const reorderLines =
+  (droppedPosition, originalPosition, lines, bookId) => (dispatch, getState) => {
```
Quiescent: Can we do the same thing as the other action here: remove `lines` and `bookId` and select those from the state.
As a general rule, we should only take a minimal amount of data as arguments.

lib/pltr/v2/selectors/secondOrder.js
```diff
@@ -212,6 +222,45 @@ export const timelineViewIsntDefaultSelector = createSelector(timelineViewSelect
...
+  pinnedPlotlinesSelector,
+  timelineViewIsStackedSelector,
+  hierarchyLevelCount,
+  selectedOrientationSelector,
+  (pinnedPlotlines, timelineViewIsStacked, hierarchyLevelCount, selectedOrientation) => {
+    if (selectedOrientation == 'vertical') {
+      return 1
+    }
+    if (timelineViewIsStacked) {
+      return hierarchyLevelCount
```
Quiescent: This still doesn't seem right.  I think you should produce something like: hierarchyLevelCount + pinnedPLotlines + 1.
Is there a reason not to?

lib/pltr/v2/actions/lines.js
```diff
@@ -42,8 +50,58 @@ export function editLineColor(id, color) {
...
+  if (line.isPinned) {
+    const reorderedLines = reorderList(pinnedPlotlines - 1, line.position, lines)
+    return dispatch({ type: UNPIN_PLOTLINE, lineId: line.id, lines: reorderedLines, bookId })
+  } else {
+    const reorderedLines = reorderList(0, line.position, lines)
+    return dispatch({ type: PIN_PLOTLINE, lineId: line.id, lines: reorderedLines, bookId })
+  }
+}
+
+export function unPinPlotline(lineId, lines, bookId) {
```
Quiescent: Please remove `lines` and `bookId` from this action.

Quiescent (changes requested):
Added a few more changes.
We need to be careful to keep the action argument lists as small as possible, otherwise it'll be a nightmare for Mobile to use them.

lib/pltr/v2/actions/lines.js
jeanawhou: @Quiescent sorry i didnt double check. Thank you for checking

jeanawhou:

lib/pltr/v2/actions/lines.js
Quiescent: @jeanawhou, we might be missing each other here.  I'm suggesting that we change the API of this function:
```
export const reorderLines =
  (droppedPosition, originalPosition, lines, bookId) => (dispatch, getState) =>
```
->
```
export const reorderLines =
  (droppedPosition, originalPosition) => (dispatch, getState) =>
```

lib/pltr/v2/selectors/secondOrder.js
```diff
@@ -212,6 +230,47 @@ export const timelineViewIsntDefaultSelector = createSelector(timelineViewSelect
...
 })
 
+const selectedOrientationSelector = (state) => state.ui.orientation
+
+export const stickyHeaderCountSelector = createSelector(
+  pinnedPlotlinesSelector,
+  timelineViewIsStackedSelector,
+  hierarchyLevelCount,
+  selectedOrientationSelector,
+  currentTimelineSelector,
```
Quiescent: This and `bookId` appear to be unused(?)

lib/pltr/v2/selectors/secondOrder.js
```diff
@@ -212,6 +230,47 @@ export const timelineViewIsntDefaultSelector = createSelector(timelineViewSelect
...
+    }
+  }
+)
+
+export const stickyLeftColumnCountSelector = createSelector(
+  pinnedPlotlinesSelector,
+  timelineViewIsStackedSelector,
+  hierarchyLevelCount,
+  selectedOrientationSelector,
+  currentTimelineSelector,
```
Quiescent: This and `bookId` appear to be unused.

lib/pltr/v2/actions/lines.js
jeanawhou: @Quiescent sorry i forgot about the reorderlines reducer 🥲

jeanawhou: Just updated the PR. Thank you for checking 🙇

Quiescent (approved):
Looks good.  LGTM.

PR #947, opened 2023-02-13
Pin plotlines

jeanawhou:
@Quiescent @cameronsutter i added unit test for both togglePlotline and reorderLines reducers


Quiescent (approved):
Good work.  Thanks for doing these 👍

PR #951, opened 2023-02-22
Bugfix/duplicate file

jeanawhou:


src/app/index.js
```diff
@@ -261,7 +261,7 @@ tellMeWhatOSImOn()
           }
         })
 
-        onSaveAs(() => {
```
Quiescent, line 264: This is here to listen to requests from the main process to save the file.
If you need a second `save-as` handler, then you need to:
 - extract the function,
 - pass the function to `onSaveAs`, and
 - use the function later for the `document.addEventListener` `save-as` handler.

Quiescent: I missed where you did exactly this.  Sorry! :)

Quiescent (approved):
Looks good.  LGTM!

